PR #357, opened 2014-11-23
Simple preview plugin that sets target="_blank" for all links when in preview mode

### LINKED ISSUE #256: Previews links are not invalidated
When editing the article, the preview is displayed through preview_inline.html. While this works fine, any internal wiki link is not (let alone links that link to the outside world), and clicking on them displays the linked page with all the navigation bars.

What if all links inside the preview were invalidated (linking to href="#")?

cXhristian:
I also added a new attribute to the Article class so that plugins can easily check if it is in preview mode or not. 

The markdown extension will ignore urls that start with # because those already work as normal. 

Fixes #256.


benjaoming:
Sorry about taking so long! This looks nice, thanks for moving the ArticleMarkdown to wiki.core.markdown, this will be an interesting module for sure and encourage that there is actually core markdown logic... which makes sense to both keep in the core and to still acknowledge that markdown is a specific language.

PR #359, opened 2014-11-25
Fix django 17

spookylukey:
Django 1.7 fixes, and other things.

Let me know if you don't want to pull all of them, I can separate out the fixes.


benjaoming:
Okay so now finally, merged!! Excitement!! I'm also resetting the south migrations.

spookylukey:
Great, thanks!

PR #362, opened 2015-01-03
Adds input type=email for wiki css

orblivion:
I noticed that creating a signup form with an email field, it isn't styled the same as password fields. I thuoght this would be useful for people moving further with adding a registration form.


benjaoming:
Thanks for this :) The change belongs in the LESS files. Would you mind putting them there? It's understandable if creating a whole LESS environment for such a small change is annoying :)

orblivion:
Hmm, yeah, and also I found that there are a handful of other changes to make for parity with other input types. There are some input['password'] ones (with the single quote). I just thought I'd do a driveby contribution but probably not worth it at this point. I'll make a github issue.

benjaoming:
You're most welcome to do drive by contributions also or issues for that matter :) Thanks!

Issue #365, opened 2015-01-05
Markdown 2.5.2 is broken?


benjaoming:
Downgrading Markdown to 2.4 fixes the following issue:

    Internal Server Error: /
    Traceback (most recent call last):
      File "/home/user/.virtualenvs/test/local/lib/python2.7/site-packages/django/core/handlers/base.py", line 99, in get_response
        resolver_match = resolver.resolve(request.path_info)
      File "/home/user/.virtualenvs/test/local/lib/python2.7/site-packages/django/core/urlresolvers.py", line 337, in resolve
        for pattern in self.url_patterns:
      File "/home/user/.virtualenvs/test/local/lib/python2.7/site-packages/django/core/urlresolvers.py", line 365, in url_patterns
        patterns = getattr(self.urlconf_module, "urlpatterns", self.urlconf_module)
      File "/home/user/.virtualenvs/test/local/lib/python2.7/site-packages/django/core/urlresolvers.py", line 360, in urlconf_module
        self._urlconf_module = import_module(self.urlconf_name)
      File "/home/user/.virtualenvs/test/local/lib/python2.7/site-packages/django/utils/importlib.py", line 40, in import_module
        __import__(name)
      File "/home/user/lalala/lalala/urls.py", line 21, in <module>
        from wiki.urls import get_pattern as get_wiki_pattern
      File "/home/user/.virtualenvs/test/local/lib/python2.7/site-packages/wiki/urls.py", line 156, in <module>
        load_wiki_plugins()
      File "/home/user/.virtualenvs/test/local/lib/python2.7/site-packages/wiki/core/plugins/loader.py", line 38, in load_wiki_plugins
        load('wiki_plugin', verbose=False, failfast=False)
      File "/home/user/.virtualenvs/test/local/lib/python2.7/site-packages/wiki/core/plugins/loader.py", line 35, in load
        get_module(app, modname, verbose, failfast)
      File "/home/user/.virtualenvs/test/local/lib/python2.7/site-packages/wiki/core/plugins/loader.py", line 17, in get_module
        module = import_module(module_name)
      File "/home/user/.virtualenvs/test/local/lib/python2.7/site-packages/django/utils/importlib.py", line 40, in import_module
        __import__(name)
      File "/home/user/.virtualenvs/test/local/lib/python2.7/site-packages/wiki/plugins/links/wiki_plugin.py", line 13, in <module>
        class LinkPlugin(BasePlugin):
[...]

PR #367, opened 2015-01-07
Properly sets default configs in plugins/links/mdx

orblivion:
See: http://pythonconquerstheuniverse.wordpress.com/2012/02/15/mutable-default-arguments/

Worth a test case?


orblivion:
In short, function default arguments are evaluated once at compile time. The same object reference will be used for each call of the object. If you use a mutable object (such as a dict) there, and it changes somewhere within the function, that changed value will now be the default when call the function the next time.

orblivion:
It looks like it is, but it should have already been fixed in October in 9d411a781713857cdc35b1decf0e239008acf656. This change just makes it a more proper fix. Looks like #365 came up just a couple days ago, but if you're using django-wiki 0.0.23 like I am, you should have the problem.

I'll put a hackish workaround in #365.

PR #368, opened 2015-01-07
Update to Bootstrap 3.3.1 and Font Awesome 4

cXhristian:
Note: I moved `@fa-font-path` to wiki.less so that the font-awesome directory doesn't contain any changes.


benjaoming:
This also closes #366

benjaoming:
massive work! I will try it out!

benjaoming:
There seems to be an issue viewing diffs, would you be able to check it out?

screenshot from 2015-01-08 00 41 14

benjaoming:
Apart from that, it looks like the upgrade and refactoring went really well. Big props!!

I think also that *should* there be any glitches, then they are cosmetical, and 0.0.24 is more about bringing a transitional release. So we can get moving!!!

benjaoming:
I'll merge this in around 24h, please add any further changes meanwhile :)

cXhristian:
Diffs seems to be broken on the demo wiki too, so I doubt this was caused by the upgrade. I can take a look at it later today though.

cXhristian:
Ok, I took a look at it. The overflow is caused by one really long line in the diff.
I guess I'll just add an overflow-x: scroll to the diff container. 

Collapse on different diffs seems a bit broken though. I'll fix that. Try opening one diff and then another.

cXhristian:
Everything should be fixed now.

PR #378, opened 2015-01-28
Use content_type instead of mimetype

cXhristian:
Mimetype was completely removed in Django 1.7


benjaoming:
Oh thanks! I've seen this somewhere before, like a deja vu of something that was already fixed once.. odd.. too many branches (there were at least two!!), nice to be back to a manageable amount :)

PR #379, opened 2015-01-29
Small refactoring of unit tests.

Alkalit:
No new asserts, only code refactoring. See names of commits.

### DIFF
--- a/wiki/tests/base.py
+++ b/wiki/tests/base.py
@@ -8,7 +8,9 @@ from wiki.models import URLPath
 class WebTestBase(TestCase):
 
     def setUp(self):
+
         super(TestCase, self).setUp()
+
         try:
             from django.contrib.auth import get_user_model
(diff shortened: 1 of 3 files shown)


benjaoming:
Thanks, appreciated! Curious, how did you get from this....

    func(
        arg1,
        arg2)

to this:

    func(
        arg1,
        arg2
    )

?

I want my autopep8 to do the same :) I actually have an issue on it.... https://github.com/hhatto/autopep8/issues/135

Alkalit:
You will be disappointed - i just made my hands dirty. It's easy with vim.
My pep8 utils also seems stupid in such cases.

benjaoming:
Oh :) but since we're gonna continue down the pep8 track, hopefully we'll end up a place where we don't need to do pep8 edits anymore.

Thanks for the contribution!

PR #380, opened 2015-01-30
view tests refactoring.

Alkalit:
And again, few test-cases splitted and moved into separate file.

Next stop is tests for template tags.


benjaoming:
Thanks, somehow Coverage also appreciates the refactoring :)

PR #382, opened 2015-02-04
Tests for template filters.

Alkalit:


benjaoming:
Thanks for this great improvement!

Would you consider using Django's own builtin decorator for purposes of changing settings? It can avoid mock as a dependency.... but to be honest, I've been part of a project where mock ruined all the tests because it was used for everything :)

https://docs.djangoproject.com/en/1.7/topics/testing/tools/#django.test.SimpleTestCase.settings

Alkalit:
Well, I'll try to rewrite.

Alkalit:
Now it's my turn to be honest - I used mock because could not achieve the same result by override decorator. For the tests I need to change the settings for the module wiki.core.permissions, but somehow built-in decorator does not work. But if in the same module to import settings directly from django.conf.settings, it has correct values. Unfortunately I do not know how to fix it.

tkliuxing:
override_settings function just send a signal `django.test.signals.setting_changed`. 
If not register callbacks function, the settings item doesn't change.

Forgive me in broken English ;-)

tkliuxing:
Maybe you can do this:
```python

from wiki.conf import settings

class CanRead(BaseTestCase):

    template = """
        {% load wiki_tags %}
        {{ article|can_read:user }}
    """
    def test_user_have_permission(self):
        settings.CAN_READ = lambda *args: True
        a = Article.objects.create()

        User = get_user_model()
        u = User.objects.create(username='Nobody', password='pass')

        output = can_read(a, u)
        self.assertTrue(output)

        output = self.render(self.template, {'article': a, 'user': u})
        self.assertIn('True', output)
```

Alkalit:
@tkliuxing thanks for assistance, but your example seems like a wheel reinventing. It breaks tests isolation, which mean for each test should be written something like this

    def setUp(self):
        super(CanRead, self).setUp()

        self.default_value = getattr(settings, 'CAN_READ', None)

    def tearDown(self):
        super(CanRead, self).tearDown()
        settings.CAN_READ = self.default_value

which can be pretty ugly and not convenient if need few settings.

Meybe in some step you will invent [context manager](http://stackoverflow.com/a/3519955/2516502). Next step is decorating.

P.S. Oh, believe me - my english even worse :D

benjaoming:
@Alkalit 
Have a look at this line at the top of the test:

    from wiki.models import URLPath

This will trigger load-time events in the whole wiki application, I think this is what breaks the settings decorator because settings are then initialized before the test is run.

Try moving all import statements interacting with `wiki` to your test methods, thereby they should be able to override settings and it also improves your isolation, I think..... looking forward to hear if you find out more about this, to be honest I never really thought about this level of test isolation.

Alkalit:
And winner in nomination workaround of the year is... ta-daaah:

    @override_settings(WIKI_CAN_READ=lambda *args: False)
    def test_user_dont_have_permission(self):

        from imp import reload
        from wiki.conf import settings
        reload(settings)

        User = get_user_model()

        a = Article.objects.create()
        u = User.objects.create(username='Noman', password='pass')

        output = can_read(a, u)
        self.assertFalse(output)

        output = self.render(self.template, {'article': a, 'user': u})
        self.assertIn('False', output)

You

Alkalit:
And second place is awarded to:

```
class wiki_override_settings(override_settings):

    def __enter__(self):
        super(wiki_override_settings, self).__enter__()

        from imp import reload
        from wiki.conf import settings
        reload(settings)
```

benjaoming:
obliged!

benjaoming:
and thanks for finding a solution to settings overrides!

Issue #387, opened 2015-02-09
Ability to override branding in page title (<title> tag)


azaghal:
Currently it is not possible to easily override the branding inside of the <title> tag (found in base_site.html).

It would be useful to be able to easily replace the default value ` - django-\/\/  i K |` with a custom one. The ` - ` should probably be enclosed with a block in order to allow people to completely remove wiki name if they wish to.

benjaoming:
This has been fixed as per above commit.

PR #388, opened 2015-02-09
Additional block for overriding branding in page title

azaghal:
Added additional block to base template that allows overriding the site title (within <title> tag). Implements #387.

It's a fairly simple patch, although I'm not sure if it would be better to allow people to be able to override the entire <title> tag in one go. I'm assuming that most people won't want to do that, though, and just want to change the branding part.

### DIFF
--- a/docs/customization.rst
+++ b/docs/customization.rst
@@ -14,18 +14,22 @@ turn simply inherits from ``wiki/base_site.html`` (adding
 nothing). ``wiki/base_site.html`` provides a complete HTML page, but provides a
 number of blocks that you might want to override. The most useful are:
 
+* ``wiki_site_title``
 * ``wiki_header_branding``
 * ``wiki_header_navlinks``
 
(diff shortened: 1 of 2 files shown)


benjaoming:
Of interest, how Wagtail uses django-overextends:

http://wagtail.readthedocs.org/en/latest/howto/custom_branding.html

And django-apptemplates:

https://pypi.python.org/pypi/django-apptemplates/

azaghal:
Hm... I can't exactly pin where the CI failed, but it seems like it couldn't download Pillow archive?

```
Downloading https://pypi.python.org/packages/source/P/Pillow/Pillow-2.7.0.zip#md5=da10ee9d0c0712c942224300c2931a1a

error: [Errno 104] Connection reset by peer
```

benjaoming:
Yeah it happens all the time these days, I think Travis has some major issues to resolve about how much they hammer their own and other's systems...

I usually just restart the jobs until the red light turns green...

PR #389, opened 2015-02-10
Tests for article model

Alkalit:
@benjaoming what do you think?

I also have are few questions:
1 URLPath looks tricky for me. Maybe simply mock this test (mean no library, just monkey patching)?
2 What does "NB" in Article methods doc's mean (I dont have formal CS degree)?

### DIFF
--- a/wiki/tests/base.py
+++ b/wiki/tests/base.py
@@ -1,3 +1,6 @@
+from __future__ import unicode_literals
+from __future__ import absolute_import
+
 from django.core.urlresolvers import reverse
 from django.test import TestCase
 from django.test.client import Client
@@ -73,6 +76,8 @@ class BaseTestCase(TestCase):
         return Template(template).render(Context(context))
 
 
+# See
+# https://github.com/django-wiki/django-wiki/pull/382
 class wiki_override_settings(override_settings):
 
     def __enter__(self):
--- a/wiki/tests/test_models.py
+++ b/wiki/tests/test_models.py
@@ -42,7 +42,7 @@ class ArticleModelTest(TestCase):
     # XXX maybe is redundant test
     def test_model_manager_class(self):
 
-        self.assertTrue(isinstance(Article.objects, ArticleManager))
+        self.assertIsInstance(Article.objects, ArticleManager)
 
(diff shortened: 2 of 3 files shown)


benjaoming:
*NB!* is not a CS degreen thing... more an old-school English thing :)

> Nota bene (/ˈnoʊtɑːˈbɛnɛ/; plural form notate bene) is an Italian and Latin phrase meaning "note well". The phrase first appeared in writing circa 1721.

But just means "take care".

URLPath, I would suggest creating a tree structure in the test and see that paths are generated and resolved correctly? At first grasp, I don't think it needs mocking, but have you encountered any issues?

benjaoming:
Btw thanks again!!

benjaoming:
Great test structure again!

The `test_get_absolute_url_if_urlpath_set_is_not_exists` test expects an article to be in '/1/', which would assume that the wiki urlconf is installed at the root level. Which could be the case, however it will not always hold.

It would be interesting (maybe more for the URLPath tests) with a test that ensures that URLs are correct both when the wiki is running from '/' and '/my-wiki/' :)

Will merge no matter what, just flinging in an idea :+1:

Alkalit:
I know what does Nota Bene means. Just do not expected to see it in the docs. I thought it had something related to databases and data structures.

No issues. Just trying to figure out how it works.

Ok, will check this, too.

Alkalit:
I got another issue. When I run testcases separatly, (say wiki.tests.test_models.ArticleModelTest.test_get_absolute_url_if_urlpath_set_is_exists__no_root_urlconf) tests passes. But when I try launch them all at once, I getting error. It looks like isolation issue again.

benjaoming:
Thanks for this! I'm looking into merging it.

I think the problem may have to do with the way that the urlconf is loaded using `wiki.urls.get_pattern()`, maybe it's only called once before the tests run... that's what I gather from debugging it now.

benjaoming:
I tried forcing a reload of the urlconf but with no success, because the already loaded module has been injected elsewhere and URLs will be loaded from `reverse()` using the already loaded urlconf.

PR #390, opened 2015-02-11
Style fixes

cXhristian:
I tried to separate all the changes into different commits as you might not want all the changes I did. 

I'm a bit unsure how wiki-control works compared to form-control so an explanation would be nice. I just added form-control on two forms, but maybe I should use wiki-control?


benjaoming:
Hi @cXhristian thanks once again for contributing and for positive improvements and making things look better and be more useful!!

Sorry that it took me so long to get this merged. But I poked around in the code, tried it out.. and it really should've been merged with the blink of an eye.

I really need to start whipping people in my org. to use the wiki more beyond the few techies, because it's becoming quite great :)

PR #392, opened 2015-02-11
Hide article menu labels in mobile view

cXhristian:
Before:

After: 

Looks the same for any size bigger than xs.

### DIFF
--- a/wiki/templates/wiki/includes/article_menu.html
+++ b/wiki/templates/wiki/includes/article_menu.html
@@ -6,7 +6,7 @@
   {% if not user.is_anonymous %}
   <a href="{% url 'wiki:settings' article_id=article.id path=urlpath.path %}">
     <span class="fa fa-wrench"></span>
-    {% trans "Settings" %}
+    <span class="hidden-xs">{% trans "Settings" %}</span>
   </a>
   {% endif %}
 </li>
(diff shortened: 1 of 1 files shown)


tkliuxing:
If only use icon, it is recommended to adjust the icon style.
I feel these icons can't express the meaning of the text.

In addition, add the title attribute is a good idea?

Forgive me in broken English ;-)

cXhristian:
Yeah, I kinda agree. 
Although I can't think of a good solution since tooltips and mobile usually work pretty bad. 

Adding a title attribute is a good idea, yeah. That should fix it on desktop.

benjaoming:
The lock icon next to "view source" is kind of important.. I think it communicates quite clearly the limited access which is the underlying reason for why editing is disabled.... also, the idea is stolen from Mediawiki :) But it could be changed into a disabled editing icon? Am not sure what other options are in the font-awesome package, but please feel free to suggest icons, also for the desktop version!

benjaoming:
I've merged it, but then we can discuss the icons separately :) The mobile version was much needed! Thanks!!

cXhristian:
Tried out some alternative icons: 





Font Awesome has support for stacking icons, but the size doesn't fit so we would have to do something custom (not super hard). 



(Imagine I was a bit less awful at design.)

benjaoming:
Would it make sense maybe to have the "source" icon in mobile view and the
lock+"view source" in larger grids? Conceptually something like:

    <span icon="fa fa-source visible-sm"></span>
    <span class="fa fa-lock visible-lg visible-md"></span>
    {% trans "View Source" %}
    <span class="hidden-xs">{% trans "View Source" %}</span>

Stacking icon might be a bit too tricky down the road, keeping custom css
in mind, too.

Russell-Jones:
I suggest

View -- document
Attachment -- paperclip
View source -- </> as above
Changes -- document overlaid with a small clock to the bottom right. The clock should have a margin to clearly separate it.
Settings -- cog as above

PR #395, opened 2015-02-14
Tests for template tags.

Alkalit:
Look closely at the WikiRenderTest. This tag performs operations with plugins, but for testing purposes I mocked this part.


benjaoming:
Thanks for this! I'm sorry that it took me so long to merge. Really glad to have more tests for django-wiki!!

PR #396, opened 2015-02-24
Fixed the compatible with the `USERNAME_FIELD` for the Django version below 1.5.

WayneSan:
The `User` model without `USERNAME_FIELD` attribute might get error in `PermissionsForm`.


benjaoming:
Thanks for this fix and sorry about the slow merge! I'm not sure that anything below Django 1.4 is actually working anymore, but please do let me know if this is relevant to you!

PR #397, opened 2015-02-25
Fixes settings.py comment

orblivion:
Assuming I got the idea right.


benjaoming:
Yes, minutes as in minutes :)

Issue #413, opened 2015-04-26
Plugin: DIsqus comment


tkliuxing:
This Plugin only add a javascript to base.html template.
It's achieved in each page can comment.
Has been achieved in this site: http://wiki.oyhy.me

![p8oaf6lwqnxegaaaabjru5erkjggg](https://cloud.githubusercontent.com/assets/652939/7336914/ea2883d8-ec45-11e4-8e7e-cc2da2a8c433.png)

benjaoming:
@tkliuxing looks nice, it's a matter of customizing your local project templates, right?

Btw. congrats on being the first 100% translated language :)

benjaoming:
Okay, so this approach will likely be described as a section in the docs on how you can use a template override to add your own Disqus comment block to all your articles.

PR #416, opened 2015-04-26
Plugin: disqus comment

tkliuxing:
I want create a plugin into wiki project, you could adopt it?
This plugin I mentioned in Issues #413 .

### DIFF
--- a/wiki/plugins/disqus/README.md
+++ b/wiki/plugins/disqus/README.md
@@ -1,0 +1,6 @@
+# HOWTO
+
+1. Register an Disqus account in here: https://disqus.com/
+2. Sign in and create a site profile: https://disqus.com/admin/create/
+3. Add `"wiki.plugins.disqus.context_processors.disqus",` to TEMPLATE_CONTEXT_PROCESSORS in settings.py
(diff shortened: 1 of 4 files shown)


benjaoming:
Hi @tkliuxing 

Congratulations on implementing Disqus.

As mentioned in #413, if people need to use Disqus, it's quite straight forward to add it to their own template override.

I do not agree that this is a "core" plugin.. but another point is that you are not really writing a pluggable plugin.

1. Your plugin affects base_site.html, a template that's OUTSIDE of the plugin
1. Your plugin uses a context processor, something that is highly discouraged as you are executing code on ALL pages of a WHOLE project.

I recommend that you use a simple template override and share the technique possibly in the documentation, that would be very useful to people and fit more nicely to conventional code patterns.

tkliuxing:
I agree with you.
Can I write it in docs/customization.rst? 

BTW: how to translate the document?

benjaoming:
I would suggest creating an additional link in tips.rst and adding a separate document in tips/disqus.rst

Regarding translation, are you sure you would want to take on such a task? I mean, the docs are only for developers, and in case you fall behind with the translation, there would be incorrect docs for Chinese, which might be worse than having English ones? Sphinx support internationalization through gettext, but I rather doubt that it's worth the effort at this stage.

PR #422, opened 2015-05-04
Add Disqus comment tips to document.

tkliuxing:


benjaoming:
Thanks! This is nice! Not only for Disqus but also because it shows a nice recipe for customizations.

Issue #418, opened 2015-04-28
Search pagination doesn't work


cXhristian:
It doesn't include q in the query string. 

Example:

URL for search: /_search/?q=a

Pagination links: /_search/?page=2

Should be: /_search/?q=a&page=2

benjaoming:
Created a new label for these nice little things :)

benjaoming:
Fixed by @csrcordeiro !

PR #429, opened 2015-05-19
Search pagination fix

csrcordeiro:
Hi everyone,
I just added what I think it's the simplest fix for the pagination problem as pointed out in issue #418. 
I tested it with the 'testproject'.


benjaoming:
Thanks @csrcordeiro that looks like a very reasonable fix :)

PR #433, opened 2015-06-15
Escape unicode filename while download attachment

bargool:
PR #384 doesn't work with apache
Got
> embedded newline in response header with name 'Content-Disposition'
From Apache

### DIFF
--- a/wiki/core/http.py
+++ b/wiki/core/http.py
@@ -7,6 +7,7 @@ from datetime import datetime
 from django.http import HttpResponse
 from django.utils.http import http_date
 from django.utils import dateformat
+from django.utils.encoding import filepath_to_uri
 
 from wiki.conf import settings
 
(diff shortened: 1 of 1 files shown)


benjaoming:
This makes sense given the old TODO, thanks for this!

bargool:
As I understand from [this SO](http://stackoverflow.com/questions/1361604/how-to-encode-utf8-filename-for-http-headers-python-django) this solution works only for modern browsers (I don't have old browsers - cannot try)

benjaoming:
Thanks @bargool for investigating, I gather that the conclusion from StackOverflow is this:

> Update: as of November 2012, all current desktop browsers support the encoding defined in RFC 6266 and RFC 5987 (Safari >= 6, IE >= 9, Chrome, Firefox, Opera, Konqueror).

benjaoming:
...which is 2½ years ago :) So I am comfortable with this solution!

bargool:
Thanks for quick response and merge!

PR #438, opened 2015-07-07
Better tox ini

spookylukey:
Speaks for itself really. Features in tox 1.8 (2014-09-24) allowed a lot of clean up.


benjaoming:
What wonderful new syntax! Thanks so much for keeping the tox updated and clean!

PR #439, opened 2015-07-08
Django 15 fixes

spookylukey:
OK, apologies in advance for a long tale...

First, I noticed that on Django 1.5 only 2 tests were running instead of 65. You can see this here:
https://travis-ci.org/django-wiki/django-wiki/jobs/66873647 (Django 1.5)
https://travis-ci.org/django-wiki/django-wiki/jobs/66873648 (Django 1.6)

The first commit fixes that.

Once the tests are running, there are test failures:
* one fairly straightforward one about striptags. This is fixed in the second changeset
* one nightmare causing an infinite recursion, and potential data loss bugs, due to `get_query_set` chaos.

This is addressed in the third changeset. Fixing it is tricky, but the pattern to use is given in the Django 1.6 release notes. https://docs.djangoproject.com/en/1.8/releases/1.6/#get-query-set-and-similar-methods-renamed-to-get-queryset . 

There are other subtleties/chaos, as detailed here: http://lukeplant.me.uk/blog/posts/handling-django%27s-get_query_set-rename-is-hard/ (see bottom for recent update). This caused a failure on Django 1.8! But this is due to `TreeManager` from `django-mptt` not following the right pattern.

I have submitted a pull request to fix `django-mptt`: https://github.com/django-mptt/django-mptt/pull/388

Until that lands and is usable, however, I've fixed the error by using a monkey patch, which is the final changeset.

I've put some of these comments in the individual commit messages, but thought the overall picture might be helpful too.

### DIFF
--- a/wiki/tests/test_template_filters.py
+++ b/wiki/tests/test_template_filters.py
@@ -145,12 +145,12 @@ class GetContentSnippet(TemplateTestCase):
 
         content = """
         <h1>Some dummy</h1> text. <div>Actually</div> I don't what to write,
-        heh. Don't now, <b>maybe</b> I should citate <>Shakespeare<> or Byron.
+        heh. Don't now, <b>maybe</b> I should citate Shakespeare or Byron.
         Or <a>maybe</a> copy paste from <a href="http://python.org">python</a>
         or django documentation. Maybe.
         """
 
-        expected = ('I should citate <>Shakespeare<> or Byron. '
+        expected = ('I should citate Shakespeare or Byron. '
             'Or <strong>maybe</strong> copy paste from python '
             'or django documentation. <strong>maybe</strong> .')
 
--- a/wiki/managers.py
+++ b/wiki/managers.py
@@ -10,6 +10,20 @@ from django import VERSION as DJANGO_VERSION
 from mptt.managers import TreeManager
 
 
+if django.VERSION >= (1, 6):
+    # TreeManager bug:
+    if 'get_query_set' in TreeManager.__dict__:
+        # TreeManager should not define this, it messes things up.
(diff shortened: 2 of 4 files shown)


wiki/tests/test_template_filters.py
```diff
@@ -145,12 +145,12 @@ def test_strip_tags(self):
 
         content = """
         <h1>Some dummy</h1> text. <div>Actually</div> I don't what to write,
-        heh. Don't now, <b>maybe</b> I should citate <>Shakespeare<> or Byron.
```
benjaoming: Isn't this supposed to test that non-HTML tags stay? Did it fail to pass?

benjaoming:
Thanks for ensuring Django 1.5 compatibility, thanks for ensuring tests of what's being fixed, and thanks for making an upstream issue for what had to be monkey patched!!

10/10 !! And thanks for the nice tale, for me that's good literature :)

Just one minor question about the test case.. see code comments

Best,
Ben

wiki/tests/test_template_filters.py
spookylukey: I'm not entirely sure what the original intention was. Having `<>` in some HTML just seems incorrect - if you want a literal `<` it should be written `&lt;`, and that's what you'd have if you entered it in an editor.

I therefore thought it was understandable that the behaviour of Django's `striptags` has changed, and probably django-wiki shouldn't have a test for this specific corner case. Don't know.

benjaoming: Agreeing! I was just wondering if there had been an original intention with this test case.

benjaoming: Thanks for explaining!

PR #449, opened 2015-07-30
Typo and missing import in installation instructions.

Russell-Jones:
Tested with Django 1.8 + Python 3.4.2

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -174,7 +174,7 @@ In Django 1.8, it should look like this::
                     'django.template.context_processors.request',
                     'django.template.context_processors.static',
                     'django.template.context_processors.tz',
-                    'django.contrib.messages.context_processors.messages'
+                    'django.contrib.messages.context_processors.messages',
                     "sekizai.context_processors.sekizai",
                 ],
             },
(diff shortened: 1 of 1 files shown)


benjaoming:
Oh thanks, I guess the intention of the urls.py thing was to ne a copy-paste snippet:

    urlpatterns += patterns

It won't work without having a previously defined `urlpatterns` anyways..

Should we instead make a fully working urls.py?

Russell-Jones:
I was thinking it'd add as a mnemonic to add the import to the default url.py (it's needed in the one for 1.8 at least) rather than people cutting and pasting, getting an error, puzzling over where to import from, then adding the import, which is how it went for me.

benjaoming:
Doesn't Django 1.8 actually raise deprecation warnings if you use the `pattern()` construct?

benjaoming:
Agreeing with the change of course :) Let's address the issue about patterns() later, it would also be erroneous to have a documentation that doesn't work for django 1.6 while it's officially supported.

Russell-Jones:
Hi, would have replied earlier if I'd not needed to restore from backup after HDD failure :)
I guess it can be updated when 1.6 drops out of support in distros, or it can be updated to use try/except for the newer form (though I don't know what that would be, TBH).

benjaoming:
Welcome back :)

I think we can leave this until 1.6 is dropped, probably won't be long...

Issue #369, opened 2015-01-15
Horizontal scrolling on code segments


benjaoming:
The CSS for `<pre>` should add horizontal scrolling and not wrap lines. Here is some test code:

    +-------+------------+-----------------------------+--------------+------------------+-----------+-------------+----------+--------+------+------------+---------+---------------+
    | Table | Non_unique | Key_name                    | Seq_in_index | Column_name      | Collation | Cardinality | Sub_part | Packed | Null | Index_type | Comment | Index_comment |
    +-------+------------+-----------------------------+--------------+------------------+-----------+-------------+----------+--------+------+------------+---------+---------------+
    | page  |          0 | PRIMARY                     |            1 | page_id          | A         |     2309270 |     NULL | NULL   |      | BTREE      |         |               |
    | page  |          0 | name_title                  |            1 | page_namespace   | A         |        NULL |     NULL | NULL   |      | BTREE      |         |               |
    | page  |          0 | name_title                  |            2 | page_title       | A         |     2309270 |     NULL | NULL   |      | BTREE      |         |               |
    | page  |          1 | page_random                 |            1 | page_random      | A         |     2309270 |     NULL | NULL   |      | BTREE      |         |               |
    | page  |          1 | page_len                    |            1 | page_len         | A         |     2309270 |     NULL | NULL   |      | BTREE      |         |               |
    | page  |          1 | page_redirect_namespace_len |            1 | page_is_redirect | A         |       24566 |     NULL | NULL   |      | BTREE      |         |               |
    | page  |          1 | page_redirect_namespace_len |            2 | page_namespace   | A         |       36655 |     NULL | NULL   |      | BTREE      |         |               |
    | page  |          1 | page_redirect_namespace_len |            3 | page_len         | A         |     2309270 |     NULL | NULL   |      | BTREE      |         |               |
    +-------+------------+-----------------------------+--------------+------------------+-----------+-------------+----------+--------+------+------------+---------+---------------+

PR #450, opened 2015-07-31
Horizontal scrolling code segments

inflrscns:
Added support for horizontal scrolling in code segments, rather than wrapping/breaking the code, as referenced in [this issue.](https://github.com/django-wiki/django-wiki/issues/369)

Here's a picture of it in action:
horizontal scrolling


benjaoming:
Thanks for finding and fixing this!

Issue #204, opened 2013-10-22
Image within a table


vincecarney:
Inserting an image within a table causes a new table row to wrap the image.

inflrscns:
I've been looking into this a little, and I noticed that this bug does not only appear when you're inserting an image within a table. Example:
```
This text [image:id] will appear as one line
```
renders as:
```
This text will appear as one line
<div class="tumbnail" style="width: 250px">
... etc
```

PR #452, opened 2015-08-01
Patch for image markdown

inflrscns:
Fixes images inside tables so they arent displayed in a new row. Images that were inserted besides text on a single line no longer appear after the text, the text will now wrap around the image.

Edit: Forgot to mention, this patch is in regards to [this issue.](https://github.com/django-wiki/django-wiki/issues/204)

## Test markdown:
test markdown  
<br/>
## How it renders before:
before  
<br/>
## And after:
after

### DIFF
--- a/wiki/plugins/images/markdown_extensions.py
+++ b/wiki/plugins/images/markdown_extensions.py
@@ -29,6 +29,8 @@ class ImagePreprocessor(markdown.preprocessors.Preprocessor):
 
     def run(self, lines):
         new_text = []
+        previous_line = ""
+        line_index = None
         previous_line_was_image = False
         image = None
         image_id = None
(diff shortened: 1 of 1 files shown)


benjaoming:
Hi @inflrscns !

Thanks for this fix! And thanks for the nice image proofs. Would you be able to screenshot the changes regarding:

 * Images with captions
 * Images that are not inline
 * Images with captions in tables

In case errors are same as before, it's not that I mind merging, we can put such problems in a separate issue.

Regards

inflrscns:
Hello @benjaoming,

Here are some additional screenshots. Let me know if you need any more, or if I missed something. I'm happy to supply them!

### Test Markup
test markup

### Results
results

PS: Speaking of separate issues, I noticed that thumbnail \<div>'s are wrapped in \<p> tags, which is not valid html. Do you think I should open an issue for this?

benjaoming:
> PS: Speaking of separate issues, I noticed that thumbnail <div>'s are wrapped in <p> tags, which is not valid html. Do you think I should open an issue for this?

Yes - please do! I have a feeling, though, that the culprit is Markdown adding the `<p>` tags by default to what it considers paragraphs... but it's probably fixable.

benjaoming:
It looks like the patch works well!! :+1:

inflrscns:
Thank you!! :)

PR #460, opened 2015-09-01
Fix errors in import script

the-glu:
This fix multiples issues in the import script for mediawiki:

* `only_printable` was broken
* Switch from `%` to `format` syntax
* Fix unicode issues with `u""` (and `format`)


cXhristian:
No need to use u'' when `unicode_literals` is imported.

the-glu:
Fixed :)

benjaoming:
Nice, thanks for modernizing the code as well! Out of interest, how did your mediawiki import go?

the-glu:
Seems to be ok, we're checking now before removing mediawiki ^^

Issue #154, opened 2013-05-20
Moving articles


benjaoming:
Everything about moving articles goes here.

First of all, django-mptt does not support concurrent CRUD: https://github.com/django-mptt/django-mptt/issues/236

benjaoming:
This topic is now so hot it has an FAQ entry :)

http://django-wiki.readthedocs.io/en/master/faq.html

benjaoming:
Move function added in #640. No links are corrected, but an option to create redirection placeholder articles is provided.

Let's continue further requests or bugs in separate issues.

PR #461, opened 2015-09-01
Move article: first version

the-glu:
This is a proposed feature to move articles in the tree of articles. It's useful eg. after a mediawiki import to organize articles :)

I don't really know if there is no side-issues to this. At least one exist: links are not updated inside others articles of the wiki and there is no redirection left.
I tested to move articles around and I didn't see any incoherence in the database nor partial issues.

I had to include CSS in a `style` tag. This is probably wrong but it's seem to be only a file for bootstrap. Where can I include extra CSS ?


benjaoming:
Hi @the-glu 

That's ambitious :) But also a feature that is often missed!!

Re: CSS: Everything is written in LESS. You need to do changes [here](https://github.com/django-wiki/django-wiki/blob/master/wiki/static/wiki/bootstrap/less/wiki/wiki.less) and run build-less.sh.

As you say, links will break if there's not a mechanism to fix it properly. Also caches will have to be cleared for article lists to work. In fact, invalidating the cache is non-trivial, and you risk to have to invalidate the whole thing.

Maybe you can respond in #154 as that's the issue where there's some discussion already taking place :)

the-glu:
Okay ! So:

- About the django-mptt/django-mptt#236 I don't think it's should be fixed in django-wiki, right ? Hopefully they will fix the problem, I don't think in a simple django-wiki installation it's an issue (aka moving a article 'alone' won't trigger it).

- I added cache invalidation and it's fix the problem with the article index (and not the TOCs who is only for the article - I assume you made a typo ;) )

- I moved the CSS to the correct part, thanks for the pointer ^^

- About redirection I totally agree. I will try to make a pull request with redirection (maybe today), and then one to add redirection when moving article, but I think this can already be merged when ready, to split issues in different parts :)

the-glu:
Ping ? :)

wiki/templates/wiki/move.html
```diff
@@ -0,0 +1,75 @@
+{% extends "wiki/article.html" %}
+{% load wiki_tags i18n sekizai_tags %}
+
+{% block wiki_pagetitle %}{% trans "Move" %}: {{ article.current_revision.title }}{% endblock %}
+
+{% block wiki_contents_tab %}
+
+{% if not user.is_authenticated %}
```
benjaoming: Is this page eligible for anonymous users? Was just wondering if it makes sense or if there's a danger or complexity in allowing that?

the-glu: It's just a warning about IP beeing looged - same as article editing.

wiki/views/article.py
```diff
@@ -401,6 +401,80 @@ def get_context_data(self, **kwargs):
...
+        kwargs = self.get_form_kwargs()
+        return form_class(**kwargs)
+
+    def get_context_data(self, **kwargs):
+
+        kwargs['root_path'] = models.URLPath.root()
+
+        return super(Move, self).get_context_data(**kwargs)
+
+    def form_valid(self, form):
```
benjaoming: Maybe this should have an atomic decorator? While the re-parenting in mptt seems fairly atomic, there's a couple of other calls surrounding it.

wiki/templates/wiki/move.html
```diff
@@ -0,0 +1,75 @@
...
+</p>
+{% endif %}
+
+<div class="row">
+
+  <div class="col-lg-8">
+    <form method="POST" class="form-horizontal" id="article_move_form">
+      <input type="hidden" name="save" value="1" />
+      <input type="hidden" name="preview" value="1" />
+      {% with form as form %}
```
benjaoming: Should this be here? And the two above fields?

wiki/templates/wiki/move.html
```diff
@@ -0,0 +1,75 @@
...
+{% block wiki_contents_tab %}
+
+{% if not user.is_authenticated %}
+<p class="alert alert-warning">
+{% trans "<strong>Warning:</strong> You are not logged in. Your IP address will be logged." %}
+<a href="{% url 'wiki:login' %}?next={% url 'wiki:move' article_id=article.id path=urlpath.path %}">{% trans "Click here to log in" %} &raquo;</a>
+</p>
+{% endif %}
+
+<div class="row">
```
benjaoming: Maybe a nice little intro about breaking links etc... mentioning the "known issues" about moving articles.

wiki/templates/wiki/move.html
the-glu: I did exactly as https://github.com/django-wiki/django-wiki/blob/master/wiki/templates/wiki/edit.html#L22 . Should something beeing changed ?

wiki/templates/wiki/move.html
benjaoming: Yes but the point was more if anonymous users are supposed to see this page?

the-glu: I can add a new right, would it be ok ?

benjaoming:
Looks good :) But could you add a test for all this? :)

the-glu:
Ok, I will correct the few remarks and try to add a test :)

wiki/templates/wiki/move.html
benjaoming: Well, this page is not for editing the article contents, so all of the above is redundant :)

benjaoming: ...and below 2 lines :)

wiki/forms.py
```diff
@@ -124,6 +124,11 @@ class CreateRootForm(forms.Form):
         required=False, widget=getEditor().get_widget())  # @UndefinedVariable
 
 
+class MoveForm(forms.Form):
+
+    destination = forms.CharField(label=_('Destination'), initial='', help_text=_('Childrens are also moved. Be carefull, links are NOT updated !'))
```
benjaoming: How is it validated that the destination is a valid slug?

wiki/views/article.py
```diff
@@ -401,6 +401,80 @@ def get_context_data(self, **kwargs):
...
+
+        dest_path = get_object_or_404(models.URLPath, pk=form.cleaned_data['destination'])
+
+        if not self.urlpath.parent:
+            messages.error(
+                self.request,
+                _('This article cannot be moved because it is a root article.'))
+            return redirect('wiki:get', article_id=self.article.id)
+
+        dest_is_children = False
```
benjaoming: dest_has_children :)

wiki/forms.py
benjaoming: I'm a bit confused -- what's this field actually used for? It seems it's for a PK for URLPath? Should it perhaps be made a ModelChoiceField with widget=TextInput() ?

wiki/views/article.py
```diff
@@ -401,6 +401,80 @@ def get_context_data(self, **kwargs):
...
+            ancestor.article.clear_cache()
+
+        self.urlpath.parent = dest_path
+        self.urlpath.save()
+
+        # Reload url path form database
+        self.urlpath = get_object_or_404(models.URLPath, pk=self.urlpath.pk)
+
+        # Use a copy of yourself (to avoid cache) and update articles links
+        # agains
```
benjaoming: Why is necessary to clear the cache one more time? Could the one before be deleted then?

ostrokach:
What is the status of this? 
Being able to move article trees around would be an awesome feature! 
Right now I'm just copy-pasting text...

benjaoming:
@ostrokach - I added the "needs-tests" label. The status is that I'm willing to merge this if there are tests written for special cases like:

 - moving an article to an already existing location
 - ensuring that moving an article does x to already existing links
 - ensuring that moving an article does x to already existing links of all child pages

There are also a couple of unanswered comments.

I don't think this PR is far from being merge-able, but I'm not able to make the effort of finalizing it.

Since I haven't heard from @the-glu for a while, I think it's probably up to others to branch out and make a new PR..

benjaoming:
Closing this PR, partly because we changed away from Coveralls, and party because keeping a very old, unmerged PR seems to imply that django-wiki has a problem accepting PRs :)

But I really hope someone - if not @the-glu - will carry this to the finish line, although I can see that it does need a bit of work to arrive.

Thanks for the work so far!

PR #462, opened 2015-09-02
[Unfinished] Globalhistory feature

the-glu:
This is a new plugin to have a page listing all changes in the wiki.

It's very basic but should be ok. The plugin may have not very good performances as he need to check rights to do pagination, instead of doing normal paginated queries in the database, but this should be ok for most of the case (and can be turn off anyways if needed).

In the `base_new.html` template I added a simple check to include the new menu only if the plugin is activated. It may be interesting to do it as well for the notification plugin, since it's seems to be a issue to turn it off, based on the documentation ^^


wiki/plugins/globalhistory/templates/wiki/plugins/globalhistory/globalhistory.html
```diff
@@ -0,0 +1,114 @@
...
+      {% if page %}
+        {% trans "No more changes to display !" %}
+        <a href="?page={{page|add:"-1"}}">{% trans "Go back to previous page" %}</a>
+      {% else %}
+        {% trans "No changes to display !" %}
+      {% endif %}
+    </div>
+  {% endif %}
+
+  <nav>
```
benjaoming: You should be able to replace this whole block by including wiki/includes/pagination.html:

https://github.com/django-wiki/django-wiki/blob/master/wiki/templates/wiki/includes/pagination.html

wiki/templates/wiki/base_site.html
```diff
@@ -66,6 +66,11 @@
                     </a>
                   </li>
                   {% include "wiki/plugins/notifications/menubaritem.html" %}
+                  {% with tpl_name="wiki/plugins/globalhistory/menubaritem.html" %}
```
benjaoming: This is an interesting way of discovering the plugin. We could also create some other sort of template tag, like a `wiki_plugin_enabled` that takes a plugin name as a string and tests if its in INSTALLED_APPS.

Because in case we do a template refactor, the above will break, which might not be very obvious in the future..

benjaoming: I'm thinking you could put it in the wiki_tags template tag library...

wiki/plugins/globalhistory/views.py
```diff
@@ -0,0 +1,58 @@
...
+        needed = (page + 1) * RESULTS_PER_PAGE
+
+        current_pos = 0
+        results = []
+        # Get all modifications from database and fill results when
+        # pos > skip and len(results) < needed
+
+        for article_revision in models.ArticleRevision.objects.order_by('-modified'):
+
+            if article_revision.article.can_read(self.request.user):
```
benjaoming: You can do this in the query set. Find the inspiration in the query set method for the Article manager:

https://github.com/django-wiki/django-wiki/blob/master/wiki/managers.py#L93

benjaoming:
This is a nice feature and certainly very useful! Wonderful!

I've added some comments.

I'm definitely willing to merge this and let it improve over time. It could do with some tests, too.

But apart from that, I'm mostly concerned with what I've commented on... the over all goal is just that the bundled plugins reflect some kind of "best practice" as far as we can go.

the-glu:
I didn't know I can use the filter to query articles based on rights, I will fix it, probably tomorrow :) 

( Then I will get a correct paginator and I can use the paginator block ;) )

About the template detection, that was a quick fix. If you want, I can add a list of template to include in the `BasePlugin` object and automatically build the list, it's seems better no ?

benjaoming:
> I can add a list of template to include in the BasePlugin object and automatically build the list

Not sure, could you explain what it would be for? If you're detecting a plugin, it might be easier to use a template tag specifically for that...

the-glu:
Yes, I but speak about listing in the pluging definition witch templates should be included, to avoid hardcoded test in the core template/templatetag.

Other possibility: try to include a predefined template based on plugins (eg. `wiki/plugins/<plugin>/menubaritem.html`). Maybe simpler.

benjaoming:
@the-glu here:

#463

the-glu:
OK, but here you add something in the code for a plugin , this is not very dynamic as core had to be modified for a new plugin. 

Wouldn't be better to try the template for each activated plugin in a loop?

benjaoming:
> OK, but here you add something in the code for a plugin , this is not very dynamic as core had to be modified for a new plugin. 

Checking if a template exists is as bad as checking if a plugin exists :) Maybe even worse in terms of runtime.

I didn't modify the core as such for a specific plugin... I did modify the template but that was to check if notifications are enabled before trying to load its menu item. I've added a template tag that you can use to explicitly check if a plugin is enabled? The idea is that you have this case:

We want to add awareness about specific plugins in our core templates.

Ideally, we specify plugin APIs for everything to make `plugin_enabled` a redundant feature. Let's do that in the future.

the-glu:
Okay :)

I switched to a ListView with a paginator, used the direct function for rights and used the new template tag for the template, everything should be ok now !

the-glu:
Ping ? :)

benjaoming:
Okay, this is looking good, but it really needs tests :) Otherwise it *will* break :)

benjaoming:
Sry bout slowness!

benjaoming:
@the-glu are you still around and would like to update the PR with a test?

the-glu:
@benjaoming I'm still around, unfortunately I haven't time to make some tests :(

I anyone want to write them it would be great, otherwise I will try to find time.

( Same for #461 )

benjaoming:
@the-glu I guess no one has shown up to work on this, just wanted to voice my encouragement and support for this PR :)

benjaoming:
Keeping this open because *It Just Needs Tests* ®

...and some merge conflicts resolved.

benjaoming:
Superseded by #627 :)

PR #465, opened 2015-09-14
Add option to use custom Group model with wiki

fritz-k:
This patch lets you use a custom `GROUP_MODEL` instead of the builtin `auth.Group`.

In a project of mine, we substituted `auth.Group` with our custom `groups.Group` model as injecting lots of functionality into `auth.Group` seemed like a big hack and it was easier this way. Unfortunately, this brings incompatibilities to 3rd party apps like this one.

This patch implements this functionality for Django 1.7 and up since it relies on `django.apps` for finding the `GROUP_MODEL` and I wasn't quite sure how to hack this change into south migrations. For Django < 1.7 it simply uses `auth.Group` despite the value of `GROUP_MODEL`.

### DIFF
--- a/wiki/conf/settings.py
+++ b/wiki/conf/settings.py
@@ -173,6 +173,9 @@ SEARCH_VIEW = getattr(
 # other objects that are changed.
 CACHE_TIMEOUT = getattr(django_settings, 'WIKI_CACHE_TIMEOUT', 600)
 
+# Choose the Group model to use. Defaults to django's auth.Group
+GROUP_MODEL = getattr(django_settings, 'WIKI_GROUP_MODEL', 'auth.Group')
+
 ###################
 # SPAM PROTECTION #
 ###################
--- a/wiki/conf/settings.py
+++ b/wiki/conf/settings.py
@@ -174,6 +174,7 @@ SEARCH_VIEW = getattr(
 CACHE_TIMEOUT = getattr(django_settings, 'WIKI_CACHE_TIMEOUT', 600)
 
 # Choose the Group model to use. Defaults to django's auth.Group
+# Has no effect in Django 1.6 and below.
 GROUP_MODEL = getattr(django_settings, 'WIKI_GROUP_MODEL', 'auth.Group')
 
 ###################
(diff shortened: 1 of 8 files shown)


wiki/conf/settings.py
```diff
@@ -173,6 +173,9 @@
 # other objects that are changed.
 CACHE_TIMEOUT = getattr(django_settings, 'WIKI_CACHE_TIMEOUT', 600)
 
+# Choose the Group model to use. Defaults to django's auth.Group
```
benjaoming: Could you add a comment here, that it's django 1.7+ only? :)

benjaoming:
This would be a great feature!

Could you add a test for this, though, since using a custom group model can easily break? I'm afraid that it has a very high risk of breaking without a proper test case.

fritz-k:
Sorry, didn't want to push that last one...

fritz-k:
Finally got it working :)
Hope those tests are ok.

wiki/tests/testdata/models.py
```diff
@@ -8,3 +8,7 @@ class CustomUser(AbstractUser):
         some_field = models.IntegerField(default=0)
 except ImportError:
     pass
+
+
+class CustomGroup(models.Model):
```
benjaoming: Wondering a bit here... does it make sense to have a custom group with no fields? I'm just a bit surprised that it actually works :)

cXhristian: In practice it doesn't make sense, but for testing it's completely fine imo.
It should act just like models.Model would do.

benjaoming: Yeah, maybe just leave it like that and then we'll discover once we impose requirements for a Group model, so we can fix the test and document it at the same time.....

@fritz-k do you feel like adding a doc section about the Group model ? :)

fritz-k: As far as I saw in the code, the Group model only needs a __str__ function for display and a ForeignKey to User for permissions, but they don't seem to get tested so this blank model should suffice until then.

fritz-k:
Where exactly should I document this? Settings, Customizations, ...?

benjaoming:
Ah, I see you've added the comment in settings.py which for now has been
the convention for all other settings, anyways.

I think it should be pretty easy to understand, so let's leave it there :)

benjaoming:
ARGH, I'm slow -- for the record, don't be over-gentle about pinging me.. just like super-ping me in case there's a stall like this. Merging this great PR with tests and comments now! Thanks @fritz-k for the contribution!

fritz-k:
Thanks for the merge. My project stalled for a while, so this fell off my radar too...

PR #471, opened 2015-10-13
Allow django-mptt > 0.7.1 as a dependency

spookylukey:


benjaoming:
Maybe django-mptt<0.8 ?

It's been breaking a lot :/

spookylukey:
Ah. Normally I would be optimistic and leave it open-ended - to me, "django-mptt<0.8" means "I know this doesn't work with 0.8". If people discover a problem with a specific mptt version they can always avoid that by choosing carefully, but if you've over-specified the version then you've potentially caused them a bigger problem.

I'm happy to change it to what you think though.

benjaoming:
I think it's fine to put upper limits.

I've had lots of problems deploying old projects with django-cms and django-blog-zinnia because if missing upper limits.

For django-mptt, the suspicion is based on experience :)

benjaoming:
**) I also had problems with these projects, because they didn't have an archived pip freeze of their virtualenvs.

spookylukey:
Updated as per suggestion now.

benjaoming:
Grrr8! Merging!

benjaoming:
Thanks and have a nice day :)

PR #474, opened 2015-10-22
Fix various deprecation warnings on Django 1.7/1.8

spookylukey:
Probably best reviewed one commit at a time, rather than the full diff.


benjaoming:
Woah, so much goodness, I really like the changes enforced by these warnings. Django truly is beautiful!!

PR #475, opened 2015-10-25
Add WIKI_BASEURL as a configuration option for sites where the wiki isn't based off /

AdamLuchjenbroers:
This commit adds an additional setting (WIKI_BASEURL) that is used to provide a base URL for the Wiki itself. This is required where the wiki is not hosted off / to allow links to work properly between pages.

This is accomplished by creating and loading the WikiLinkExtension for Markdown, passing it the provided base URL path.


AdamLuchjenbroers:
Hmm, might hold off on this one, as it doesn't appear to handle the hierarchy within the Wiki properly.

Issue #466, opened 2015-09-23
Saving an article without title returns error


Gheoan:
* Django 1.8.4
* Python 3.4.3
* Windows 10
* How to reproduce: 
 1. Go to an already created article.
 2. Click Edit.
 3. Clear the title field
 4. Save
* Using django-wiki from master branch

benjaoming:
```

  File "./wiki/forms.py", line 198, in clean
    if cd['title'] == self.initial_revision.title and cd[
```

PR #476, opened 2015-10-27
Validate revision title

thomastu:
This is in regards to issue #466. 

This PR changes the `EditForm` clean behaviour to throw a validation error when attempting to save something with no title, or with an all whitespace title.

### DIFF
--- a/wiki/forms.py
+++ b/wiki/forms.py
@@ -188,6 +188,11 @@ class EditForm(forms.Form, SpamProtectionMixin):
         super(EditForm, self).__init__(*args, **kwargs)
 
     def clean(self):
+        """Validates form data by checking for the following
+        No new revisions have been created since user attempted to edit 
+        Revision has a title 
+        A title has 
+        """
         cd = self.cleaned_data
         if self.no_clean or self.preview:
             return cd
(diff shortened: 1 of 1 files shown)


wiki/forms.py
```diff
@@ -195,6 +200,9 @@ def clean(self):
             raise forms.ValidationError(
                 ugettext(
                     'While you were editing, someone else changed the revision. Your contents have been automatically merged with the new contents. Please review the text below.'))
+        if ('title' not in cd) or (not cd['title'].strip()):
```
benjaoming: I would suggest adding a `clean_title` method instead since that will prompt the error message to show up next to the title field.

```python
def clean_title(self):
    title = self.cleaned_data.get('title')
    title = (title or "").strip()
    if not title:
        raise forms.ValidationError(ugettext('Article is missing title or has an invalid title'))
    return title

benjaoming:
Thanks for the fix :) And thanks for adding in an extra strip(), that seems useful! Added a comment about moving the validation, would you mind this change?

benjaoming:
Looks good, thanks for this :+1:

PR #487, opened 2015-10-31
Fix diff and merge AJAX call server errors

### LINKED ISSUE #486: AJAX calls to revision diff and merge gives 500 server error
When viewing changes, clicking the version links gives 500 error, as the call to the diff() view function fails.
The same goes for merging of different versions.

The reason being that attaching a function pointer to a class is basically the same as
assigning a new instance method to the class. Thus, the first method argument is 'self', which
in this case is the URLPatterns object itself, when it should be the Django request object.

A PR is coming up shortly.

myth:
Fix a bug introduced in 0e3d363dcdc39167d652bcd1fe44d838df131cef where the function pointers to diff and merge view functions are attached as an instance method on the class instead of a staticmethod. This caused 'self' to be passed as first argument to these views, resulting in stacktraces for these views.

This was new to me, but attaching a function outside of the constructor of the class renders
that function to be declared as an instance method.

Performing the same operation inside the `__init__(self)` constructor will automatically yield a staticmethod.

Examples:

The following will produce an instancemethod:

```
def func(req, *args, **kargs):
    pass

class MyClass(object):
    some_view = func
```

Where a call to MyClass().some_view will have the following parameters: `(self, req, *args, **kwargs)`

However, if this is done inside the constructor, func will automatically be decorated with `@staticmethod`:

```
from module import func

class MyClass(object):
    def __init__(self):
        self.my_view = func
```

MyClass().my_view() is now a static method.

Until the `diff` and `merge` views are converted to class-based, the `diff` and `merge` view functions needs to be decorated with `staticmethod()` to prevent passing along the object reference upon invocation.

Hope you can get this merged soon because my inbox is hammered with stacktraces triggered by our dear friend googlebot.

Cheers

Fixes #486

### DIFF
--- a/wiki/urls.py
+++ b/wiki/urls.py
@@ -32,11 +32,11 @@ class WikiURLPatterns(object):
     article_settings_view_class = article.Settings
     article_source_view_class = article.Source
     article_plugin_view_class = article.Plugin
-    revision_change_view = article.ChangeRevisionView
-    revision_merge_view = article.merge
+    revision_change_view_class = article.ChangeRevisionView
+    revision_merge_view = staticmethod(article.merge)
 
     search_view_class = settings.SEARCH_VIEW
-    article_diff_view = article.diff
+    article_diff_view = staticmethod(article.diff)
 
     # account views
     signup_view_class = accounts.Signup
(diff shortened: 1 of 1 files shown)


cXhristian:
Good job!

benjaoming:
That makes sense, seems we miss tests for these cases.

PR #489, opened 2015-11-01
Update installation.rst

reduxionist:
Add trailing '+' after version number on comment indicating that 'django.contrib.sites' should be added to INSTALLED_APPS for Django versions 1.6.2(+)


benjaoming:
Thanks for noticing and fixing, that makes sense!

PR #492, opened 2015-11-11
Update settings.py

steffann:
Add missing WIKI_ prefix, with backward compatibility for old name


benjaoming:
@steffann --  agree to add the WIKI_ prefix

Tests are failing because of a syntax err, could you

fix the syntax error and indentation
git commit --amend
git push -f

Issue #498, opened 2015-12-03
Installing from github fails if pypandoc already installed


tbabej:
First make sure pandoc is available:

    $ sudo pip install pypandoc
    Collecting pypandoc
      Downloading pypandoc-1.1.2.zip
    ...
    Successfully installed pypandoc-1.1.2

Then, trying to install django wiki from git:

    $ pip install git+git://github.com/django-wiki/django-wiki
    Collecting git+git://github.com/django-wiki/django-wiki
      Cloning git://github.com/django-wiki/django-wiki to /tmp/pip-73D6bn-build
        Complete output from command python setup.py egg_info:
        Traceback (most recent call last):
          File "<string>", line 20, in <module>
          File "/tmp/pip-73D6bn-build/setup.py", line 77, in <module>
            long_description = pypandoc.convert(get_path('README.md'), 'rst')
          File "/usr/lib/python2.7/site-packages/pypandoc/__init__.py", line 50, in convert
            outputfile=outputfile, filters=filters)
          File "/usr/lib/python2.7/site-packages/pypandoc/__init__.py", line 68, in _convert
            raise RuntimeError('Missing format!')
        RuntimeError: Missing format!
    
        ----------------------------------------
    Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-73D6bn-build

benjaoming:
Pypandoc was there because README.rst was auto-generated. But since I've changed that in #493, it can be completely removed from setup.py -- and should be!

tbabej:
Right. I updated the pull request.

benjaoming:
Does it work now? :)

tbabej:
Yes, issue is fixed, thanks.

PR #499, opened 2015-12-03
setup: Make sure README conversion does not terminate the setup

tbabej:
This solves #498.


benjaoming:
Thanks for looking into this, luckily it was even easier than the fix :) A couple of weeks ago, removing README.rst was motivated by a similar issue, but I just forgot to clean up setup.py after

PR #519, opened 2016-01-08
"Are you sure you want to leave this page" confirmation for Edit page.

inflrscns:
When a user tries to leave the edit page with unsaved changes, a dialogue box will pop up asking for confirmation. 

I also updated every instance of the deprecated "alert-error" class to "alert-danger".


benjaoming:
Thanks for this!

PR #521, opened 2016-01-12
Undo weird old changes to exception handling

cXhristian:
I have absolutely no clue what the old code was trying to do. 
The relevant commit (791888eec2d64c80bb3abd0e43fbfcdb1b91a18a) seems to just update the code to the new exception syntax except for these two places. 

What it ended up doing was overwriting DoesNotExcept with an instance of another DoesNotExist. The next time this exception was supposed to be called it failed and created another exception (TypeError: 'DoesNotExist' object is not callable).


benjaoming:
Nice catch!

PR #531, opened 2016-02-29
Fix login page title

irvind:
I have noticed a weird title on the login page. There is my tiny fix.


benjaoming:
What a strange mistake... by me! If only I could travel back in time... 3 years.. I would slap myself... https://github.com/django-wiki/django-wiki/commit/45cd25e65225b629437f48e2d0aa6532907bf0ec

Thanks for fixing it! There's been an update in django-nyt for Dj 1.9 compatibility so I might as well get them both updated and released...

Issue #490, opened 2015-11-01
Changing password (or user account page)


jluttine:
I may have missed something, but I can't find a link to a page where a user could change his/her password. And maybe some other user account settings like email address. Should it be a part of django-wiki by default or do admins need to create this feature themselves?

benjaoming:
Correct, account handling has never been an intended feature. However, the password change might be so common, useful, and even warranted for good security practices that we should have it? You are welcome to post a PR for this feature -- remember a test in that case :)

jluttine:
Or would a plugin be better?

benjaoming:
Hmm, that would be a big refactor... I'm about to go on leave and finish 0.1, so it would be bad timing.. but I would really welcome a "password change" page within the current structure if you're able to work on it.

jluttine:
Ok! I probably don't have the time for it now but I'll keep this in mind when I'll be working again on my wiki site project next time.

y0no:
It will be really great to include this feature  :+1:

benjaoming:
Coming up in next release, thanks to @inflrscns for the work!

PR #535, opened 2016-03-09
account settings page

inflrscns:
https://github.com/django-wiki/django-wiki/issues/490
Adds a simple page where users can change their password and email address. It looks like this:
screenshot

I also fixed a small typo ("ArticlManagerTests" -> "ArticleManagerTests")

### DIFF
--- a/wiki/templates/wiki/base_site.html
+++ b/wiki/templates/wiki/base_site.html
@@ -65,12 +65,14 @@
                       {% trans "Log out" %}
                     </a>
                   </li>
+                  {% if "ACCOUNT_HANDLING"|wiki_settings %}
                   <li>
-                    <a href="{% url 'wiki:update' %}">
+                    <a href="{% url 'wiki:profile_update' %}">
                       <i class="fa fa-gear"></i>
                       {% trans "Account Settings" %}
                     </a>
                   </li>
+                  {% endif %}
                   {% if user.is_superuser %}
                   <li>
                     <a href="{% url 'wiki:deleted_list' %}">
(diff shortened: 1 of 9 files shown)


wiki/templates/wiki/base_site.html
```diff
@@ -65,6 +65,12 @@
                       {% trans "Log out" %}
                     </a>
                   </li>
+                  <li>
+                    <a href="{% url 'wiki:update' %}">
```
benjaoming: The url name "update" is a little too generic :)

wiki/urls.py
```diff
@@ -42,6 +42,7 @@ class WikiURLPatterns(object):
     signup_view_class = accounts.Signup
     login_view_class = accounts.Login
     logout_view_class = accounts.Logout
+    settings_view_class = accounts.Update
```
benjaoming: "settings_view_class" ? Could also be "profile_update_view_class"

benjaoming:
Thanks! We could definitely use a bit more "batteries included" by having this option for users to change their passwords.

I think that the link shouldn't be shown when account handling is off. So the template should have a condition such that it can check the value of settings.ACCOUNT_HANDLING

A couple of comments above, too, and I did a PR for your code regarding making a proper barrier when account handling is off.

benjaoming:
Could you write a small test, too?

inflrscns:
Okay, I think that should cover everything. Let me know if it needs any more changes!

benjaoming:
Awesome work, I somehow got mislead to think there was something missing in this PR. I had notifications from Github off for a long time due to offline-like conditions.

So, going to merge this. Might fix a couple of pep8-like things, but this looks great nonetheless! Thanks! Will come out in next release!

PR #539, opened 2016-04-28
good url demo in readme

tiregram:
i think the url demo  in the readme is false.


benjaoming:
Correct! Thanks! The old one used to work, too, but was reconfigured some days ago when I decided to add SSL to the demo.. because why not!? :D

Issue #287, opened 2014-07-19
Pep8


benjaoming:
When there is:

 * Just 1 branch
 * No pending pull requests

We shall autopep8 the codebase!

PR #541, opened 2016-05-05
Adding precommit hooks and removing unused imports, fixing pep8 etc

benjaoming:
Pep8 has slightly degraded since the cleanup. I think it's warranted with pre-commit hooks.

#287 #276


benjaoming:
Coverage falls even though 105 LOC was removed!?

Issue #533, opened 2016-03-03
IntegrityError at /info/_plugin/images/delete/1/


andrew66:
IntegrityError at /info/_plugin/images/delete/1/
(1048, "Column 'revision_number' cannot be null")
Request Method:	GET
Request URL:	http://127.0.0.1:8000/info/_plugin/images/delete/1/
Django Version:	1.9.3
Exception Type:	IntegrityError
Exception Value:	
(1048, "Column 'revision_number' cannot be null")
Exception Location:	/Users/longfei/develop/virtualenv/wikienv/lib/python3.5/site-packages/MySQLdb/connections.py in query, line 280
Python Executable:	/Users/longfei/develop/virtualenv/wikienv/bin/python
Python Version:	3.5.1
Python Path:	
['/Users/longfei/develop/pyprojects/wikiproj',
 '/Users/longfei/develop/pyprojects/wikiproj',
 '/Users/longfei/develop/virtualenv/wikienv/lib/python35.zip',
 '/Users/longfei/develop/virtualenv/wikienv/lib/python3.5',
 '/Users/longfei/develop/virtualenv/wikienv/lib/python3.5/plat-darwin',
 '/Users/longfei/develop/virtualenv/wikienv/lib/python3.5/lib-dynload',
 '/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5',
 '/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/plat-darwin',
 '/Users/longfei/develop/virtualenv/wikienv/lib/python3.5/site-packages',
 '/Users/longfei/develop/pyprojects/wikiproj/wikiproj',
 '/Users/longfei/develop/pyprojects/wikiproj']
Server time:	Thu, 3 Mar 2016 22:03:21 +0800

benjaoming:
@andrew66 we need a full traceback to debug. Unfortunately, the fact that the error occurred in MySQLdb does not hint to much.

Am trying to investigate it, anyways.

benjaoming:
There we go! Thanks for reporting - a traceback proving your point would probably have accelerated things a bit ;)

benjaoming:
Released in 0.1.2

PR #547, opened 2016-05-10
Restore image restoring

benjaoming:
Adds:

 * Seemingly stubbed out (unimplemented feature) of setting the image revision finalized

Fixed:

 * Deleting and restoring images in general was borked because the revision number wasn't set on the new revision #533
 * Deleting and restoring images with current revision unset


benjaoming:
Dunno why @coveralls isn't submitting a new coverage report comment. It seems to have received the data: https://coveralls.io/builds/6119628

PR #550, opened 2016-06-03
Fixed typo.

blag:


benjaoming:
A bit loose with the O's there it seems :) Thanks! :)

PR #552, opened 2016-06-19
Responsive tables

### LINKED ISSUE #403: Tables aren't responsive
Bootstrap has a .table-responsive class that tables can be wrapped in to enable horizontal scrolling on small devices for big tables. 

http://getbootstrap.com/css/#tables-responsive

duvholt:
Fixes #403

Wrap all markdown tables with `<div class="table-responsive">`.
I included a simple test.


wiki/core/markdown/mdx/responsivetable.py
```diff
@@ -0,0 +1,41 @@
...
+class ResponsiveTableExtension(markdown.Extension):
+    """Wraps all tables with Bootstrap's table-responsive class"""
+
+    def extendMarkdown(self, md, md_globals):
+        md.treeprocessors.add('responsivetable', ResponsiveTableTree(md), "_end")
+
+
+class ResponsiveTableTree(Treeprocessor):
+    def run(self, root):
+        for table_wrapper in list(root.getiterator('table')):
```
duvholt: I couldn't get this to work without wrapping the iterator in a list. Otherwise it iterated forever even with one element.

benjaoming: That's fair, no reason to expect table objects to fill up the memory :)

wiki/core/markdown/mdx/responsivetable.py
```diff
@@ -0,0 +1,41 @@
...
+            table_new = self.create_table_element()
+            self.convert_to_wrapper(table_wrapper)
+            self.move_children(table_wrapper, table_new)
```
benjaoming: Does this somehow leave an empty `<table>` element?

duvholt: I can't think of a case where it would. 

The logic is like this: 

- Create new table element
- Move children (thead, tbody) from old table to new table
- Append new table to old table
- Rename old table tag to div and add Bootstrap class

It's done this way to preserve parent and children of the table.
There might be better ways to add elements in the middle of an ElementTree, but I couldn't find any.

benjaoming:
Really great work!!

Let me know if there's anything you want to discuss about this, because at first look, I can't really spot anything apart from that one small thing I've commented on.

wiki/core/markdown/mdx/responsivetable.py
benjaoming: Ah okay, so the existing `<table>` becomes a `<div>` and then a new `<table>` element is created and inserted?

duvholt: Yep!

benjaoming:
Merging this... brings our total commit count over 1,500 :)

PR #559, opened 2016-07-30
URL supports numbers  "dates"

teury:


benjaoming:
In principle, it would be nice to have a complete separation of top-level slugs and article ID, but at this point, I'm not in favor of breaking the URL scheme. There are other ways to fix the issue, for instance by not allowing pure numbers in the database.

For other PRs, in case you're interested: The patch would also have to come with test updates, release notes and a proper description.

PR #565, opened 2016-09-11
Fix markdown extensions memory leak

duvholt:


wiki/core/markdown/__init__.py
```diff
@@ -22,7 +22,7 @@ def core_extensions(self):
 
     def get_markdown_extensions(self):
         kwargs = settings.MARKDOWN_KWARGS
-        extensions = kwargs.get('extensions', [])
+        extensions = list(kwargs.get('extensions', []))
```
benjaoming: Not sure if I'm understanding the issue fully -- but shouldn't it be a `set()` rather than a list to avoid duplicate entries?

benjaoming:
Thanks for looking into this quickly! :)

wiki/core/markdown/__init__.py
duvholt: No, `list()` is used to make a copy of the extension list. 
Basically done to avoid modifying the original markdown extension list.

benjaoming: Cool! I also like the test, it proves that you are right of course :)

PR #570, opened 2016-09-26
wiki.core doesn't have settings.py; In fact wiki.conf does

nehemiahjacob:
Addressing:

File "/Users/neo/.virtualenvs/pp_new/lib/python3.5/site-packages/wiki/plugins/haystack/__init__.py", line 3, in <module>
    from wiki.core import settings
ImportError: cannot import name 'settings'

Process finished with exit code 1

### DIFF
--- a/wiki/plugins/haystack/__init__.py
+++ b/wiki/plugins/haystack/__init__.py
@@ -1,6 +1,6 @@
 from __future__ import absolute_import
 from __future__ import unicode_literals
-from wiki.core import settings
+from wiki.conf import settings
 from wiki.core import permissions
 from wiki import models
 


benjaoming:
Ah, thanks!

I'm not actively using the Haystack plugin, could you indicate if it's working after this change?

nehemiahjacob:
It does; It might need few other trivial changes; I will keep em posted in a week; I do see this commit has not been delivered to PyPi. What is the release cycle? :)

benjaoming:
@nehemiahjacob - well I'm going to do something this week - if you have time, could you put a PR with a release notes entry about your fix? (in `docs/release_notes.rst`)

PR #582, opened 2016-11-22
Fix duplicate search results

duvholt:
If searching for articles while logged in, but without permissions to edit the article, the search returns duplicates.  More specifically it returns a duplicate for each member of the group the article is owned by.

This is because of the query produced by the many-to-many relation using `LEFT OUTER JOIN`. 
I read in an old issue that distinct was deliberately omitted because of performance issues so I only added it to the one place I know there is a bug.
Might be smarter to add it in `can_read` instead, but I'm not sure where the method is used and if it would lead to performance issues.

### DIFF
--- a/wiki/managers.py
+++ b/wiki/managers.py
@@ -43,7 +43,7 @@ class ArticleQuerySet(QuerySet):
             q = self.filter(Q(other_read=True) |
                             Q(owner=user) |
                             (Q(group__user=user) & Q(group_read=True))
-                            )
+                            ).distinct()
         return q
 
     def can_write(self, user):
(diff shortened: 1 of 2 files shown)


benjaoming:
Nice find!

I think it should be implemented directly in `can_read`, I can't think of a reason why any place should want duplicates, and as you say, a JOIN is used in a manner, that will produce duplicates. So it's warranted to filter them out.

benjaoming:
@duvholt Happy Holidays :)

Would you be able to move the distinct filtering to the `can_read` filter or would you have an opinion otherwise?

duvholt:
@benjaoming happy holidays to you too!

Completely forgot to update this pr. I pushed the change now!

benjaoming:
That's awesome thanks - adding release note in #586 :)

Issue #554, opened 2016-06-19
Slug field js error: ReferenceError: downcode is not defined


benjaoming:
When creating new articles, the slug field is no longer auto-populated. 

Django 1.9

teury:
Error # slug  https://demo.django.wiki/0/

if you put a single number in the slug creates an error

benjaoming:
Opened up #558 for that.

benjaoming:
Fixed in #591

PR #591, opened 2016-12-27
Fix slug validation behavior

### LINKED ISSUE #558: Slug validation should deny numbers-only
@teury wrote:

> if you put a single number in the slug creates an error

...meaning that the form accepts '1' as a slug, but the url pattern doesn't match. The point is that a number is interpreted as an Article ID and not a slug path, so any pure number shouldn't be accepted as a slug.

benjaoming:
Fixes #558 and #554

### DIFF
--- a/wiki/static/wiki/js/urlify.js
+++ b/wiki/static/wiki/js/urlify.js
@@ -1,0 +1,185 @@
+/*
+Patched version of Django's URLify since Django put everything into a protected
+namespace and hence we couldn't change its behavior from outside.
+*/
+/*global XRegExp*/
+(function() {
+    'use strict';
+
+    var LATIN_MAP = {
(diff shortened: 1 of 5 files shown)


wiki/static/wiki/js/urlify.js
```diff
@@ -0,0 +1,185 @@
...
+        */
+        var removelist = [];
+
+        var r = new RegExp('\\b(' + removelist.join('|') + ')\\b', 'gi');
+        s = s.replace(r, '');
+        // if downcode doesn't hit, the char will be stripped here
+        if (allowUnicode) {
+            // Keep Unicode letters including both lowercase and uppercase
+            // characters, whitespace, and dash; remove other characters.
+            s = XRegExp.replace(s, XRegExp('[^-_\\p{L}\\p{N}\\s]', 'g'), '');
```
nschneid: Where is `XRegExp` defined? I'm trying to use this code in a custom form and am getting an error.

nschneid: Oh I see, I need to include the script from /static/admin/js/vendor/xregexp/xregexp.min.js to use `allowUnicode=true`.

PR #595, opened 2017-01-10
Lowercase slugs when creating new pages with [[Like This]].

ericclack:
Currently in django-wiki slugs must be lowercase. You can create a wiki link [[Like This]] in any page content, and the system redirects you to the create form if you click it (which is expected behaviour). However, the slug field is set to "Like_This", which fails validation because slugs must be lowercase.

This change lowercases the slug before the form loads, so that the user can proceed to add their new page without validation error.

Auto test added to `wiki/tests/test_views.py`

### DIFF
--- a/wiki/decorators.py
+++ b/wiki/decorators.py
@@ -105,7 +105,7 @@ def get_article(func=None, can_read=True, can_write=False,  # noqa
                     return HttpResponseRedirect(
                         reverse(
                             "wiki:create", kwargs={'path': parent.path, }) +
-                        "?slug=%s" % pathlist[-1])
+                        "?slug=%s" % pathlist[-1].lower())
                 except models.URLPath.DoesNotExist:
                     return HttpResponseNotFound(
                         render_to_string(
--- a/wiki/tests/test_views.py
+++ b/wiki/tests/test_views.py
@@ -58,10 +58,18 @@ class ArticleViewViewTests(ArticleWebTestBase):
 
     def test_redirects_to_create_if_the_slug_is_unknown(self):
 
-        response = self.get_by_path('Unknown/')
+        response = self.get_by_path('unknown/')
         self.assertRedirects(
             response,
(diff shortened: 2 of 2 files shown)


benjaoming:
Hi @ericclack !

Thanks for the PR! It's understandable that slugs should be pre-filled as lowercase or just forced to be lowercased. This would seem more sane. We could alter the `[[Wiki Link]]` syntax to be syntactical sugar for something like [[wiki-link]]`

Would you care to comment on the retention of the underscore `_` ?

I mean, the nicest looking slugs are the ones with dashes and those are also auto-generated by the JS script.

I'm not so interested in the whole `[[Wiki Page]]` syntax since it doesn't know about hierarchy and article IDs, so the behavior or problems related to it is not *that* sensitive IMO.

benjaoming:
Thanks also for adding the test!

ericclack:
Hi Ben, 
Regarding the underscore vs hyphens, I agree hyphens look better, but I think this behaviour is controlled by the Markdown plugin `wikilinks`, see the code here:

ENV/lib/python2.7/site-packages/markdown/extensions/wikilinks.py

```
def build_url(label, base, end):
    """ Build a url from the label, a base, and an end. """
    clean_label = re.sub(r'([ ]+_)|(_[ ]+)|([ ]+)', '_', label)
    return '%s%s%s' % (base, clean_label, end)
```

So it might not be that easy to change?

benjaoming:
No, I agree, this would mean building a different extension with the
idea of django-wiki slugs and hierarchy. Which is essentially why we
created

[Title of link](wiki:ArticleSlug)


I know it takes a bit more typing, but it's aligned with normal Markdown
standards and available as a shortcut for the editor.


Best,
Ben

ericclack:
So I think we can merge in my change then, as it's independent from the underscore vs hyphen issue?

Let me know what you think...
-Eric.

benjaoming:
Absolutely, thanks for the contribution!

PR #597, opened 2017-01-30
Remove mediawikiimport

benjaoming:
0% code coverage, ~does not support Python 3~, does not support Django 1.9+, rather angry guy on mailing list, so let's get rid of it.

Sorry @the-glu - we can maybe re-add it, if it's fixed and tests are added.


the-glu:
It seems to be django-related (probably a new version with a different management of options - I never used them except here  so I don't know exactly) rather than python 3, but ok :)

benjaoming:
Ah yes, you're right, mailing list guy was complaining about something that doesn't work in Python 3 without saying what.

Issue #596, opened 2017-01-30
Bleach converts ">" to "&gt;" breaking blockquotes


nickfsolly:
Blockquotes are no longer working as bleach converts `>` to `&gt;`

We can't do a simple find/replace as not all `>` characters represent a blockquote.

Thanks all.

nickfsolly:
This issue also affects Markdown email links in the form `<joe@bloggs.com>`

benjaoming:
Yes, you're right.

Googling this, I find that everybody bleaches *after* parsing with python-markdown. My problem when implementing it, was that I couldn't reliably figure out which tags to whitelist from the Markdown-generated contents.

benjaoming:
@nickfsolly not sure how pressing your use case is, but would you be able to give feedback regarding the tag and attribute whitelist proposed in #598 ?

nickfsolly:
@benjaoming had a look at #598 and, for my usage, it looks great.  Thanks for your support.

benjaoming:
Fixed in #598, released in 0.2.1

PR #598, opened 2017-02-14
Bleach after markdown processing + First initial whitelist of tags and attributes

benjaoming:
This is an attempt of fixing #596.

I've tried with Pygments, and it works with the previous odd display of linenumbers still showing up. Should probably test some more in regards to syntax highlighting.

Also needs testing in relation to attachments.


benjaoming:
Working on the Pygments highlighting as well, now appears like this @azaghal 

The scrolling, however, does not show up at least in my Firefox test. I think the `overflow: auto/scroll` CSS property doesn't trigger on tables.

screenshot from 2017-02-15 01-00-45

azaghal:
No, that overflow setting does not apply on tables. Did some in-browser to get it working, but it looks like either the display property of tables has to be messed with, or table has to be wrapped in a `<div>` (and then you set-up the div for scrollbar). It could be this kind of fix would be needed on python-markdown side, though.

azaghal:
Hm... I do believe the thing would need changes in python-markdown for the additional div wrapper around table - possibly somewhere in `codehilite.py:123`, or even better there should be combination of changes in pygments (specify additional wrapper around table) + python-markdown (expose the option from pygments).

azaghal:
Hm... I do believe the thing would need changes in python-markdown for the additional div wrapper around table - possibly somewhere in `codehilite.py:123`, or even better there should be combination of changes in pygments (specify additional wrapper around table) + python-markdown (expose the option from pygments).

benjaoming:
This is how code highlighting looks after the changes:

screenshot from 2017-02-19 17-50-39

benjaoming:
Merging this after positive feedback from @nickfsolly

PR #600, opened 2017-03-04
replacing render_to_request with render

acer9997:


benjaoming:
Thanks for this! I guess we are talking about an area of the code that isn't covered by tests?

I can see that `render_to_response` is also used in `wiki/views/article.py`

PR #606, opened 2017-03-19
Replace render_to_request() with render ()

benjaoming:
Supersedes  #600


benjaoming:
Well okay it's not actually deprecated yet, Django 1.10 docs say: `  It’s not recommended and is likely to be deprecated in the future.`

Issue #601, opened 2017-03-05
Code block that uses backticks with numbering has inconsitent formatting and highlighting compared to regular code blocks


azaghal:
When creating code blocks in wiki pages, and Pygments library is available, code blocks are highlighted in slightly different way, with line numbering being shown to the left of each line.

A recent changeset has fixed some rendering issues related to such output, but the styling and formatting is not consistent with regular code blocks. This happens mainly with back-tick style code block syntax, although some inconsistencies exist with regular (4-space indent) code block syntax as well.

Reproduction steps:

1. Set-up Django Wiki, making sure that Pygments library is available.

2. Create page with the following content:

<pre>
This results in line-numbered code block:

```
#!/bin/bash

echo "This is one really big line that is going to be so long that it will not be able to fit into the margins of the window, and horisontal scrollbar will be needed."
echo 1
echo 1
echo 1
echo 1
echo 1
echo 1
echo 1
echo 1
echo 1
echo 1
echo 1
echo 1
echo 1
echo 1
echo 1
echo 1
echo 1
echo 1
echo 1
echo 1
echo 1
echo 1
echo 1
echo 1
```

This results in regular line-numbered code.

```
echo "This is one really big line that is going to be so long that it will not be able to fit into the margins of the window, and horisontal scrollbar will be needed."
echo 1
echo 1
echo 1
echo 1
echo 1
echo 1
```
</pre>

Expected results:

1. Line-numbered code block is same width on-screen as regular code block and has a horisontal scroll-bar.

2. Line-numbered code block has same background colour as regular code block.

3. Line-numbered code block is the same height as regular code block and has a vertical scrollbar.

Actual results:

1. Line-numbered code block is not limited in width, making the HTML page wide enough so that scrollbars will kick in on the page itself (instead of only the code block). This can be particularly annoying on mobile views (see attached screenshots).

![backtick_code_block-styling_and_formatting-full_browser_width](https://cloud.githubusercontent.com/assets/809269/23589161/b0d3aca8-01c8-11e7-8b6d-4cd890d91dc3.png)
![backtick_code_block-styling_and_formatting-mobile_browser_width](https://cloud.githubusercontent.com/assets/809269/23589162/b0d5577e-01c8-11e7-9b4c-71f19723c28b.png)

2. Line-numbered code block does not have the same background colour as the regular one - it maintains the default page colour instead of grey.

[...]

PR #608, opened 2017-04-02
Styling fix for code blocks with line numbers

azaghal:
As of current master, code blocks with line numbers (with Pygments library installed) end-up looking different compared to plain code blocks. In addition to standing out styling-wise, they also break the responsiveness of the pages by overflowing outside of visible viewport when screen is too small.

Don't hesitate to ask for improvement to the pull request, since I am not really a master of CSS.


azaghal:
Hm, forgot to reference a relevant issue, but basically this would fix issue #601.

benjaoming:
@azaghal great! do you need any additional verification or does it _just work_? :)

azaghal:
Well, works for me (tm), but would be good if you checked locally as well. And as long as the CSS looks fine to you :)

benjaoming:
Working nicely in FF 52 and Chromium :)

screenshot from 2017-04-06 13-54-34

PR #609, opened 2017-04-05
Every instantiation of ArticleMarkdown makes settings.MARKDOWN_KWARGS grow

obtroston:
Python 3.6.0, Django 1.11.

In ArticleMarkdown.__init__ the result of a call to get_markdown_extensions is written directly to settings.MARKDOWN_KWARGS, that is simultaneously used by get_markdown_extensions as a source, hence settings.MARKDOWN_KWARGS grows with every instantiation of ArticleMarkdown.

### DIFF
--- a/wiki/core/markdown/__init__.py
+++ b/wiki/core/markdown/__init__.py
@@ -13,8 +13,7 @@ from wiki.core.plugins import registry as plugin_registry
 class ArticleMarkdown(markdown.Markdown):
 
     def __init__(self, article, preview=False, *args, **kwargs):
-        kwargs = settings.MARKDOWN_KWARGS
-        kwargs['extensions'] = self.get_markdown_extensions()
+        kwargs = {'extensions': self.get_markdown_extensions()}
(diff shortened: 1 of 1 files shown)


benjaoming:
> the result of a call to get_markdown_extensions is written directly to settings.MARKDOWN_KWARGS

Where do you see this happening?

wiki/core/markdown/__init__.py
```diff
@@ -13,8 +13,7 @@
 class ArticleMarkdown(markdown.Markdown):
 
     def __init__(self, article, preview=False, *args, **kwargs):
-        kwargs = settings.MARKDOWN_KWARGS
-        kwargs['extensions'] = self.get_markdown_extensions()
+        kwargs = {'extensions': self.get_markdown_extensions()}
```
benjaoming: This removes any kwargs from the caller... I'll make a second PR, thanks for reporting this!!

benjaoming:
Thanks for finding and proposing a fix - replaced with another PR

PR #610, opened 2017-04-05
Added option to disable html sanitizer.

mhozza:


benjaoming:
Heh, fed up with it? :)

benjaoming:
I would warn that this is pretty unsafe because Markdown has removed `safe_mode` and we're not using that option anymore.

From Markdown docs:

> Both safe_mode and the associated html_replacement_text keywords are deprecated in version 2.6 and will raise a DeprecationWarning. The safe_mode and html_replacement_text keywords will be ignored in version 2.7. The so-called “safe mode” was never actually “safe” which has resulted in many people having a false sense of security when using it. As an alternative, the developers of Python-Markdown recommend that any untrusted content be passed through an HTML sanitizer (like Bleach) after being converted to HTML by markdown.

mhozza:
Actually we had `safe_mode` explicitly disabled. We wanted to be able to add any html to the articles.

benjaoming:
Oh right, well it's going to be released next time, which should be soon... adding it to my TODO now..

PR #611, opened 2017-04-06
Fix memory leak reported by @obtroston in #609

benjaoming:
Fix memory leak reported by @obtroston in #609


wiki/core/markdown/__init__.py
```diff
@@ -13,7 +13,7 @@
 class ArticleMarkdown(markdown.Markdown):
 
     def __init__(self, article, preview=False, *args, **kwargs):
-        kwargs = settings.MARKDOWN_KWARGS
+        kwargs.update(settings.MARKDOWN_KWARGS)
```
benjaoming: By changing this line, we are now working on a fresh dict on every call - and we update it with the default contents of MARKDOWN_KWARGS.

PR #612, opened 2017-04-08
Added new image size option

floemker:
Added a new option to specify the thumbnail size of images.
This patch is based on the work by inflrscns:

    https://github.com/django-wiki/django-wiki/compare/master...inflrscns:thumbnail-sizes

Extensions compared to that version are:
  - New image size "orig". No thumbnail is generated in this case, but the original image is shown instead.
  - Updated image help in the sidebar.

The size can be specified by "size:small | medium | large | orig | default".
The small/medium/large/default sizes are configurable.

### DIFF
--- a/wiki/plugins/images/markdown_extensions.py
+++ b/wiki/plugins/images/markdown_extensions.py
@@ -8,7 +8,7 @@ from django.template.loader import render_to_string
 from wiki.plugins.images import models, settings
 
 IMAGE_RE = re.compile(
-    r'.*(\[image\:(?P<id>\d+)(\s+align\:(?P<align>right|left))?(\s+size\:(?P<size>default|small|medium|large))?\s*\]).*',
+    r'.*(\[image\:(?P<id>\d+)(\s+align\:(?P<align>right|left))?(\s+size\:(?P<size>default|small|medium|large|orig))?\s*\]).*',
     re.IGNORECASE)
 
 class ImageExtension(markdown.Extension):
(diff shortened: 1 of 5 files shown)


benjaoming:
Thanks for this! Have done a manual test, and it looks really good!

Curious: How did you choose the default sizes?

benjaoming:
And sorry that the coverage tests are failing like this - there are no tests of rendering of the `[image]` markdown tag. If we had a test like that, code would pass.

floemker:
The default sizes are the ones from

    https://github.com/django-wiki/django-wiki/compare/master...inflrscns:thumbnail-sizes

No changes there till now. Do you prefer different ones? That would be an easy chance.
I will try to come up with a test.

benjaoming:
Hi @floemker !

No preferences from me - I can see that Mediawiki people have worked with this without seeming to reach a conclusion: https://www.mediawiki.org/wiki/Requests_for_comment/Standardized_thumbnails_sizes

benjaoming:
@floemker - but otherwise I'm willing to merge this now, it's a really nice feature. The fact that tests are missing are of cause a flaw, but can't say that it's a blocker, we can also add that at a later point in time ?

floemker:
I added now some initial tests. Thumbnail generation itself is not tested, but the HTML
part should be mostly covered.

Not sure if everything is correct. Quite a bit of guessing was involved.

benjaoming:
@floemker 
> Quite a bit of guessing was involved.

I can't see that, looks like nice code, and systematic tests :) :+1: 

Big thanks for adding these tests!

PR #613, opened 2017-04-08
Allow the HTML tag <br>.

floemker:
When adding two spaces at the end of lines markdown creates a \<br\> tag.
\<br\> needs to be allowed for that.


benjaoming:
That makes absolute sense, thanks for spotting this!

PR #620, opened 2017-04-15
Bootstrap upgrade from 3.3.1 to 3.3.7

benjaoming:
Routine upgrade, removed some unused CSS files from upstream Bootstrap since we only use our own custom-built version.


benjaoming:
I'm supposing the slight risk of upgrading Bootstrap will be covered by the fact that hundreds of issues have been closed in Bootstrap since 3.3.1.

PR #626, opened 2017-04-16
Adds a functional test for the TOC plugin

benwah:
Just adding a test for TOC. I think the TOC codepath is already executed in some of the tests, but this adds a TOC-specific content assertion.

I want to add a horizontal TOC markdown plugin so I need to starts with some basic tests.

ping @benjaoming

### DIFF
--- a/wiki/tests/test_markdown_plugins.py
+++ b/wiki/tests/test_markdown_plugins.py
@@ -9,6 +9,7 @@ from wiki.core.markdown.mdx.codehilite import WikiCodeHiliteExtension
 from wiki.core.markdown.mdx.responsivetable import ResponsiveTableExtension
 from wiki.models import URLPath
 from wiki.plugins.links.mdx.djangowikilinks import WikiPathExtension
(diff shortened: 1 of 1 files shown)


benwah:
Oh actually it adds to coverage - nice

benwah:
* fixed typo in test name

benjaoming:
This is surely needed! Thanks for adding!

Would you mind creating a `tests/` package in `wiki/plugins/macros` and having it sit there as `test_toc.py`? Because that would make it the start of some much needed tests for the `macros` plugin where `toc` lives..

benwah:
Sure @benjaoming I'll try to get to it tonight.

benjaoming:
Will refactor tests at a later occasion - think we should move everything to a separate `/tests/` structure as most projects do nowadays..

benjaoming:
Thanks for having such impact on test coverage @benwah ! :)

benwah:
Sounds good - I'll spin by #django-wiki on IRC at some point so we can chat how to structure tests.

PR #628, opened 2017-04-19
New bootstrapped image insert dialog

floemker:
The current image insert procedure has different shortcomings.
- Does not allow to set the image size
- Does not check/help the user when setting the alignment
- Does not check the given alignment
- Two dialogs to get two values

This implementation replaces that with one bootstrapped dialog, which
allows to select the alignment and size via drop-down menus, and
creates always a valid and minimal tag (assuming of cause that the
caption is correct markdown).

### DIFF
--- a/wiki/plugins/images/templates/wiki/plugins/images/sidebar.html
+++ b/wiki/plugins/images/templates/wiki/plugins/images/sidebar.html
@@ -6,7 +6,7 @@
 function insert_image(data) {
     if (typeof data == "number") {
         image_id = data;
-        $("#img_title").text(" {% trans "Insert Image" %} " + image_id);
+        $("#img_title").text(" {{ _("Insert Image")|escapejs }} " + image_id);
         $("#img_modal").on('shown.bs.modal', function(){
             $(this).find('#img_caption').focus();
         });
--- a/wiki/plugins/images/templates/wiki/plugins/images/sidebar.html
+++ b/wiki/plugins/images/templates/wiki/plugins/images/sidebar.html
@@ -3,15 +3,35 @@
 
 {% addtoblock "js" %}
 <script type="text/javascript">
-function insert_image(image_id) {
-  align=prompt('Enter an alignment. \'left\' or \'right\':', 'right');
-  caption=prompt('Enter a caption text. You may use markdown. You can edit the text after...');
-  imagetag = '\n[image:'+image_id+' align:'+align+']';
-  if (! caption == '') {
-    $('#id_content').insertAtCaret(imagetag+'\n    '+caption+'\n\n');
-  } else {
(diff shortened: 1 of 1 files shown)


wiki/plugins/images/templates/wiki/plugins/images/sidebar.html
```diff
@@ -3,15 +3,35 @@
...
-  imagetag = '\n[image:'+image_id+' align:'+align+']';
-  if (! caption == '') {
-    $('#id_content').insertAtCaret(imagetag+'\n    '+caption+'\n\n');
-  } else {
-    $('#id_content').insertAtCaret(imagetag+'\n\n');
-  }
+function insert_image(data) {
+    if (typeof data == "number") {
+        image_id = data;
+        $("#img_title").text(" {% trans "Insert Image" %} " + image_id);
```
benjaoming: nitpicking - when you do translation inside JS, because some languages will have weird characters, it should be `{{ _("Insert Image")|escapejs }}`. It'll be picked up by `makemessages` as well :)

wiki/plugins/images/templates/wiki/plugins/images/sidebar.html
```diff
@@ -119,10 +139,49 @@
   {% trans "How to use images" %}
 </h4>
 
-<p>{% trans "After uploading an image, it is attached to this particular article and can be used only here. Other users may replace the image, but older versions are kept. To show the image press the Insert button and fill in the caption field and possibly choose to have your image floating right or left of the content. You can use markdown in the caption. The markdown code syntax for images looks like this, possible values for align are left | right and for size are small | medium | large | orig | default:" %}<br />
+<p>{% trans "After uploading an image, it is attached to this particular article and can be used only here. Other users may replace the image, but older versions are kept. To show the image press the Insert button and select the options you want to use. You can use Markdown in the caption. The markdown code syntax for images looks like this, possible values for align are left | right and for size are small | medium | large | orig | default:" %}<br />
```
benjaoming: Could we do something to make this more readable like adding `<code>` around the syntax items?

wiki/plugins/images/templates/wiki/plugins/images/sidebar.html
floemker: Should be fixed.

wiki/plugins/images/templates/wiki/plugins/images/sidebar.html
floemker: Used now \<pre> for the possible options. The text is longer this way, but easier to read and the possible options stand out.

benjaoming:
Awesome :+1: 

I have no further comments, merging and looking forward to seeing this in action!!

PR #629, opened 2017-04-20
Allow the HTML tag <hr>.

floemker:
When placing three hyphens, asterisks, or underscores on a line by
themselves markdown creates a \<hr\> tag. \<hr\> needs to be allowed
for that.


benjaoming:
Thanks!! :+1:

PR #631, opened 2017-04-21
Refactor: tests to a separated layout, wiki package in src/

benjaoming:
Refactor:

 * All test code moved to `tests/`
 * All of the `wiki` package now lives in `src/`
 * Redid some pep8 because some of the codebase was unlinted

From Pytest's *Good Integration Practices*:

> This layout prevents a lot of common pitfalls and has many benefits, which are better explained in [this excellent blog post by Ionel Cristian Mărieș](https://blog.ionelmc.ro/2014/05/25/python-packaging).


benjaoming:
This already caught one error, having had test+development artifacts in the dist tarball/whl on Pypi.

It's definitely going to give a much better guarantee that gaps between local development environments and distributed environments don't get confused in the tests: We test the dist, not the dev environment, as the blog post "Packaging a python library" says.

src/wiki/plugins/haystack/__init__.py
```diff
@@ -64,7 +64,6 @@ def __call__(self, request):
                 models.URLPath.root().article,
                 self.request.user):
             self.results = self.__filter_can_read(self.request.user)
-            #self.results = self.results.filter(current_revision__deleted=False)
```
benwah: giphy

src/wiki/plugins/images/templatetags/wiki_thumbnails.py
```diff
@@ -0,0 +1,2 @@
+from __future__ import absolute_import
```
benwah: Why is this file needed?

tests/core/test_basic.py
```diff
@@ -7,8 +7,8 @@
 from wiki.forms import Group
 from wiki.models import URLPath
 
-from .base import wiki_override_settings
-from .testdata.models import CustomGroup
+from ..base import wiki_override_settings
```
benwah: 👍

src/wiki/plugins/images/templatetags/wiki_thumbnails.py
benjaoming: As of now, it's needed because a lot of HTML files say `{% load wiki_thumbnails %}` but I don't know if adding the `register` from sorl actually makes any difference.. I should try deleting it, you're right!

benjaoming: Just verified it, this *is* actually the reason :) I think it's to provide a decent wrapper for sorl.thumbnail on case it changes API or we add an alternative. Or maybe it's there for some historic reason.

benjaoming:
Gonna get this merged now and not let it sit in the way of future PRs!

PR #632, opened 2017-04-25
Update installation.rst

MALANGBEIJING:


MALANGBEIJING:
I use django 1.10.7

docs/installation.rst
```diff
@@ -237,15 +237,8 @@ end of your urlconf. You can also put it in */wiki* by putting
...
-        if settings.DEBUG:
-            urlpatterns += staticfiles_urlpatterns()
-            urlpatterns += patterns('',
-                                    url(r'^media/(?P<path>.*)$',
-                                        'django.views.static.serve',
-                                        {'document_root': settings.MEDIA_ROOT,
-                                         }),
-                                    )
+    urls.py
+        urlpatterns += static(settings.MEDIA_URL, document_root=settings.MEDIA_ROOT)
```
benjaoming: The block below `manually::` has to be indented, but I can see your point - the pattern is outdated for newer versions of Django, and since support is being cut to Django 1.8 (or even more restrictive, that's pending a decision), we can clean up these instructions...

benjaoming:
Thanks for this! I guess we need to update the docs for the 0.2 series in general.. installation instructions for `<Django1.8` can be removed completely.

benjaoming:
Included in #636, thanks for noticing the outdated installation instructions! They could be substantially cut down thanks to modern Django/Python technology :D

PR #624, opened 2017-04-15
[WIP] Bump Django requirement to 1.11

benjaoming:
Adding Django 1.11 support in prep for 0.3

Fixes #614

Issue #614, opened 2017-04-11
Add Django 1.11 support for 0.3 series


is161523:
Hey,
there were some changes in django.forms.widgets in django 1.11. Looks like there is something broken now.

```
Traceback:

Environment:


Request Method: GET
Request URL: http://10.10.6.53:8080/_edit/

Django Version: 1.11
Python Version: 2.7.12
Installed Applications:
['django.contrib.admin',
 'django.contrib.auth',
 'django.contrib.contenttypes',
 'django.contrib.sessions',
 'django.contrib.messages',
 'django.contrib.staticfiles',
 'django.contrib.sites',
 'django.contrib.humanize',
 'django_nyt',
 'mptt',
 'sekizai',
 'sorl.thumbnail',
 'wiki',
 'wiki.plugins.attachments',
 'wiki.plugins.notifications',
 'wiki.plugins.images',
 'wiki.plugins.macros']
Installed Middleware:
['django.middleware.security.SecurityMiddleware',
 'django.contrib.sessions.middleware.SessionMiddleware',
 'django.middleware.common.CommonMiddleware',
 'django.middleware.csrf.CsrfViewMiddleware',
 'django.contrib.auth.middleware.AuthenticationMiddleware',
 'django.contrib.messages.middleware.MessageMiddleware',
 'django.middleware.clickjacking.XFrameOptionsMiddleware']


Template error:
In template /usr/local/lib/python2.7/dist-packages/wiki/templates/wiki/includes/formfield.html, error at line 14
   build_attrs() got an unexpected keyword argument 'name'   4 : 
   5 : <div id="div_{{ field.auto_id }}" class="form-group{% if field.errors %} has-error{% endif %}">
   6 :   <div class="wiki-label col-xs-3 col-lg-2">
   7 :   {% if field.label %}
   8 :     <label for="{{ field.id_for_label }}" class="control-label {% if field.field.required %}requiredField{% endif %}">
   9 :     {{ field.label|safe }}
   10 :     </label>
   11 :   {% endif %}
   12 :   </div>
   13 :   <div class="wiki-control col-xs-9 col-lg-10">
   14 :      {{ field }}  {% if field.field.required %}<span class="asteriskField">*</span>{% endif %}
   15 :     {% if field.errors %}
   16 :       {% for error in field.errors %}
   17 :         <div id="error_{{ forloop.counter }}_{{ field.auto_id }}" class="help-block"><strong>{{ error }}</strong></div>
[...]

PR #634, opened 2017-05-02
Django 1.11 fixes

spookylukey:
This is a PR for Django 1.11 support, refs #614, using the approach described in
https://github.com/django-wiki/django-wiki/pull/624#issuecomment-298289218

I didn't start from #624 just because it was simpler to start from a clean sheet, and I had already done so.

Some additional notes, beyond what is the commit messages:

1. I did a fair bit of refactoring of test classes. This was basically because you have test base classes being used for two purposes - one is essentially fixture creation, the other is different types of test (Django client, WebTest, Selenium). To avoid combinatorial explosion of base classes, now most fixtures are provided by mixins which can be added to any test.

   I also had to fix a fair few fragile tests.

2. I had to rename the existing 'WebTestBase' to 'DjangoClientTestBase', otherwise there is just confusion due to two meanings of 'WebTest'.

3. The appearance of the edit page is now different - the 'action bar' containing delete, preview and save buttons no longer spans the whole width. This is due to how the HTML needs to nest, but it also makes more logical sense, since the forms in the side bar are actually separate and submit separately from the main one.

4. The edit page is now significantly different under the hood, though almost the same in terms of how it looks behaves. It turns out a lot of the Javascript was unnecessary, using things like 'formtarget' and 'formaction' - see http://caniuse.com/#feat=form-submit-attributes and https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button .  All the inline 'onclick' style Javascript is gone.

5. I added a utility for running tests. With the increase in options available, this makes the options much easier to find.

6. Selenium tests are off by default. But I ran them on all environments and everything passes.

7. I only added Chrome tests, and not Firefox. It's easy to add both if required, or have a switch, but Firefox is trickier, for reasons described here: http://django-functest.readthedocs.io/en/latest/installation.html#dependencies

8. There is a lot more that could be done with using django-functest. Most of the view tests could be rewritten to be clearer, shorter and less dependent on internals. We could also do proper testing of the 'concurrent editing' situation, using http://django-functest.readthedocs.io/en/latest/common.html#django_functest.FuncCommonApi.new_browser_session (for either WebTest or Selenium tests, or both).

   However, all these I left for a later day. I only rewrote the tests for creating the root article, and the tests for editing an article - and, of course, finally fixed the bug with Django 1.11!

### DIFF
--- a/src/wiki/templates/wiki/edit.html
+++ b/src/wiki/templates/wiki/edit.html
@@ -25,7 +25,10 @@
       <div class="form-group form-actions">
         <div class="col-lg-8">
           <div class="col-lg-offset-2">
-            <button class="btn btn-default" onclick="$('#previewModal').modal('show'); document.getElementById('article_edit_form').target='previewWindow'; document.getElementById('article_edit_form').action='{% url 'wiki:preview' path=urlpath.path article_id=article.id %}'; $('#article_edit_form').submit()" type="submit">
+            <button class="btn btn-default" type="submit" name="preview" value="1" id="id_preview"
+                    formaction="{% url 'wiki:preview' path=urlpath.path article_id=article.id %}"
+                    formtarget="previewWindow"
+                    >
               <span class="fa fa-eye"></span>
               {% trans "Preview" %}
             </button>
@@ -68,7 +71,7 @@
           <span class="fa fa-arrow-circle-left"></span>
           {% trans "Back to editor" %}
         </a>
-        <a class="btn btn-lg btn-primary" onclick="window.onbeforeunload = null; document.getElementById('article_edit_form').target=''; document.getElementById('article_edit_form').action='{% url 'wiki:edit' path=urlpath.path article_id=article.id %}'; document.getElementById('article_edit_form').submit();" href="#">
+        <a class="btn btn-lg btn-primary" id="id_preview_save_changes" href="#">
           <span class="fa fa-check"></span>
           {% trans "Save changes" %}
         </a>
@@ -91,7 +94,15 @@
     $("#article_edit_form").on("submit", function () {
         window.onbeforeunload = null;
         return true;
-    })
+    });
+    $("#id_preview").on("click", function () {
+        $("#previewModal").modal("show");
+        return true;
+    });
+    $("#id_preview_save_changes").on("click", function (ev) {
+        ev.preventDefault();
+        $("#id_save").trigger("click");
+    });
   });
 
 var confirmOnPageExit = function (e) {
--- a/src/wiki/templates/wiki/edit.html
+++ b/src/wiki/templates/wiki/edit.html
@@ -17,11 +17,32 @@
 
   <div class="col-lg-8">
     <form method="POST" class="form-horizontal" id="article_edit_form">
-      <input type="hidden" name="save" value="1" />
       <input type="hidden" name="preview" value="1" />
       {% with edit_form as form %}
         {% include "wiki/includes/editor.html" %}
       {% endwith %}
+
+      <div class="form-group form-actions">
+        <div class="col-lg-8">
+          <div class="col-lg-offset-2">
+            <button class="btn btn-default" onclick="$('#previewModal').modal('show'); document.getElementById('article_edit_form').target='previewWindow'; document.getElementById('article_edit_form').action='{% url 'wiki:preview' path=urlpath.path article_id=article.id %}'; $('#article_edit_form').submit()" type="submit">
+              <span class="fa fa-eye"></span>
+              {% trans "Preview" %}
+            </button>
+            <button class="btn btn-primary" type="submit" name="save" value="1" id="id_save">
+              <span class="fa fa-check"></span>
+              {% trans "Save changes" %}
+            </button>
+            {% if article|can_delete:user %}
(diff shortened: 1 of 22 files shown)


spookylukey:
Ah, the tests are failing because even though I've set it to skip the Selenium tests, the `setUpClass` method still gets called, and that is causing the error with chromedriver. I'll work on a fix.

src/wiki/templates/wiki/edit.html
```diff
@@ -17,11 +17,34 @@
...
+            {% if article|can_delete:user %}
+              <a href="{% url 'wiki:delete' path=urlpath.path article_id=article.id %}" class="btn btn-danger pull-right">
+                <span class="fa fa-trash-o"></span>
+                {% trans "Delete article" %}
+              </a>
+            {% endif %}
+          </div>
+        </div>
+      </div>
+
```
benjaoming: This is from some years back - but I think the initial solution has been related to problems in getting the correct values to be sent together with the form.

But I can't find a proper source confirming that there are - for instance - browsers that don't send `value` together with the clicked submit button.

Furthermore, because the `Preview` button is the first button, that should also be the default trigger ENTER key codes.

So it should be fine :) Am glad to see JS be deleted :dancing_men:

tests/base.py
```diff
@@ -39,47 +41,84 @@ def setUp(self):
         )
 
 
-class ArticleTestBase(TestCase):
+class RequireBasicData(RequireSuperuserMixin):
```
benjaoming: Thanks for taking time to improve naming! Agree :+1:

tests/base.py
```diff
@@ -39,47 +41,84 @@ def setUp(self):
...
+class WebTestBase(WebTestCommonMixin, django_functest.FuncWebTestMixin, TestCase):
+    pass
+
+
+INCLUDE_SELENIUM_TESTS = os.environ.get('INCLUDE_SELENIUM_TESTS', '0') == '1'
+
+
+@unittest.skipIf(not INCLUDE_SELENIUM_TESTS, "Skipping Selenium tests")
+class SeleniumBase(WebTestCommonMixin, django_functest.FuncSeleniumMixin, StaticLiveServerTestCase):
+    driver_name = "Chrome"
```
benjaoming: Guessing at some point this will be made configurable, but there's no need at present.

benjaoming:
Great PR @spookylukey - nice to get Django 1.11 and at the same time upgrade the test patterns and conventions around the codebase!

What's the status or roadmap for `DjangoClientTestBase` after this? Should it be completely factored out and replaced by `WebTest` or does it still have a reason-to-be?

benjaoming:
...my last question is a follow-up of your comment in #624 :

> My initial tests suggest I can add the django-functest FuncWebTestMixin into an appropriate test base class without having to rewrite all the tests that use Django Client, which may be the best compromise at the moment.

src/wiki/templates/wiki/edit.html
spookylukey: I remember now problems with `button` in IE6/IE7 - but according to [this](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button) they have been fixed since IE8, so only a fraction of a percent of users are affected now.

spookylukey:
Regarding `DjangoClientTestBase`, I decided in the end it was cleaner to separate out the tests that used the Django client. Otherwise people might have the expectation that the Django `Client` instance will 'know' about the WebTest interactions - when in reality they will have different sets of cookies and sessions etc.

For the future, I'm not sure what should be done. When I looked at the tests, it seemed there were a few tests that are better done using the Django client. This would probably apply especially to views that just return JSON and things like that e.g. the `diff` view. So, I don't think we should try to get rid of it entirely. I think the best approach would be to say:

* For new tests of views, use the django-functest method if possible
* When changing existing tests that use DjangoClientTestBase, or fixing/changing functionality, consider rewriting the tests to use django-functest.

Not sure where in the docs this should go, perhaps we need to organise all development related things  from the README and CONTRIBUTING.

benjaoming:
> Not sure where in the docs this should go, perhaps we need to organise all development related things from the README and CONTRIBUTING.

Yes, in lack of better options, it could go here: http://django-wiki.readthedocs.io/en/master/#contributing

..I've added #635 to address this.

benjaoming:
Thanks for explaining @spookylukey - seems like a good idea to completely isolate the use of `DjangoClientTestBase` and `WebTest` because of costly (likely+wrongful) assumptions that they share sessions+cookies. Putting it in the docs wouldn't harm, it's fine to have it sitting in `README.rst` and them write a better Developer Guide at a later occasion.

benjaoming:
> Selenium tests are off by default. But I ran them on all environments and everything passes.

Coverage FOMO... how much coverage do the new Selenium tests result in?

spookylukey:
@benjaoming I'm not sure how to measure the coverage specifically of the Selenium tests. A quick test suggests it doesn't increase Python coverage at all, which is actually what I'd hope - only Javascript coverage is increased, the Python is testable using WebTest tests.

benjaoming:
Thanks for writing a thorough explanation @spookylukey ! I'll see what happens once developer documentation has its own section and **Tests** have their own chapter. Might wanna put a bit more emphasis on this part:

> Selenium tests are not run by default

But really great work, and prosperous to find you once take test practices in a django app up a notch! :+1:

Issue #576, opened 2016-10-27
slug utf 8


teury:
it would be a good idea, for a site in other languages.

benjaoming:
It's not in the definition of what a slug is to be UTF-8 or is there something that I misunderstand?

https://en.wikipedia.org/wiki/Semantic_URL#Slug

teury:
in some language it is essential.
example:

English ----- Spanish
boy       =        Niño   
Dad      =        Papá
potato  =      Papa


Wikipedia in other languages ​​used in the url utf8.
https://es.wikipedia.org/wiki/Niño

django enables this feature from 1.9

teury:
If someone wants to have utf 8, in the url, do this.

1. Wiki/Models/urlpath.py
On line 80: add "allow_unicode=True"
`slug = models.SlugField(allow_unicode=True, verbose_name=_('slug'), null=True, blank=True,
                            max_length=SLUG_MAX_LENGTH)`

2. Wiki/forms.py
On line 317: add "allow_unicode=True"
`slug = forms.SlugField(
        allow_unicode=True,
        label=_('Slug'),
        help_text=_(
            "This will be the address where your article can be found. Use only alphanumeric characters and - or _. Note that you cannot change the slug after creating the article."),
        max_length=models.URLPath.SLUG_MAX_LENGTH)`

3. Wiki/views/article.py
On line 74: remove attrs={}

Remove:  `attrs={
                # Make patterns force lowercase if we are case insensitive to bless the user with a
                # bit of strictness, anyways
                'pattern': '[a-z0-9_-]+' if not settings.URL_CASE_SENSITIVE else '[a-zA-Z0-9_-]+',
                'title': 'Lowercase letters, numbers, hyphens and underscores' if not settings.URL_CASE_SENSITIVE else 'Letters, numbers, hyphens and underscores',
            }`

**Solution For: Django 1.9=>**

benjaoming:
@teury it's not really recommended to change the source code in deployments unless you want your project to be difficult to maintain :)

The above change will not work with the current `urlify.js` which automatically replaces well-known foreign characters with their conventional ascii replacements. For instance 'æ' becomes 'ae'. You could try out 0.2 and see if the automatic replacements are what you need.

In order to have `allow_unicode` in the slug field, we can use `allowUnicode=true` when calling `URLify`. So the change would require:

1. To be configurable with a setting
1. Using the setting for `CreateForm.slug`
[...]

PR #640, opened 2017-05-20
Move articles

floemker:
Updated and extended version of the pull request #461 "Move article: first version".

The changes compared to the last version in #461 are:

- Updated to trunk / current Django version.
- Added option to create redirect pages for all moved articles.
- Allow changing the destination slug.
  Check as well if the slug is valid and does not already exist.
- Added tests.
- Allow moving of articles only if the user is logged in.
- Moved the 'Move article' button next to the remaining buttons on the editor page.
- Added a 'Go back' button on the move page, and changed the layout of that
  page so that it's more similar to the delete article page in regard to the
  button layout.
- Name the 'Move article' button always identical.
- Typo / grammar fixes.
- Don't allow selecting any child of the to be moved article.
  Fail with an error if a page should be moved to be a child of it's own.
- Don't show the move button for the root article.
- Small minor clean up's / simplifications


benjaoming:
Great work!!

2 small suggestions, please feel free to disagree..

## Suggestion 1

> Children are also moved.

This text could be made dependent on number of children:

 * 0: Don't display text (I think that it's okay not to be explicit that there are no children, because there aren't so moving the article doesn't come with this sort of warning). Just display: "Be carefull, links to this article will not be updated".
 * >=1: `{{ article.urlpath.get_descendants.count }} nested articles will also be moved. Be carefull, links to this article and {{ article.urlpath.get_descendants.count }} articles nested in this hierarchy will not be updated.`

## Suggestion 2

I think we should have `moved` data stored in the database. Some approaches, do you have any ideas or preferences?

 * `URLPath.moved` boolean, simply indicating that the article is now at a new location
 * `URLPath.new_path` FK pointing to the new URLPath.
 * A new table URLPathRedirects or likewise, storing `url_path` and `new_destination` or similar, which will be able to map out a history of moving around articles.

## Final thoughts

I think the user flow is great... it's cool to have the "Move article" button in the edit section, it's easy to notice, and commonly moving an article will take place in the beginning stages (User creates article -> Oops wrong slug).

I'll do a line-by-line review any time :)

This is also a really big feature, so we should definitely celebrate it with a release :)

src/wiki/templates/wiki/move.html
```diff
@@ -0,0 +1,58 @@
...
+        <a class="btn btn-default dropdown-toggle" data-toggle="dropdown" href="#">
+          <span class="dest_selector_title"></span>
+          <span class="caret"></span>
+        </a>
+        <ul class="dropdown-menu" role="menu" aria-labelledby="dLabel">
+            {% with current_path=root_path %}{% include "wiki/includes/move_tree.html" %}{% endwith %}
+        </ul>
+      </div>
+      <p class="col-lg-offset-2">
+        {% trans "Children are also moved. Be carefull, links are NOT updated!" %}
```
nschneid: typo: should be spelled "careful"

nschneid:
These sound like nice improvements. Are slugs containing non-ASCII letters still prohibited? This renders them unusable for many languages.

src/wiki/templates/wiki/move.html
floemker: Should be fixed.

floemker:
Just committed Suggestion 1.1.

floemker:
> Suggestion 2
> 
> I think we should have moved data stored in the database. Some approaches, do you have any ideas or preferences?
> 
>     URLPath.moved boolean, simply indicating that the article is now at a new location
>     URLPath.new_path FK pointing to the new URLPath.
>     A new table URLPathRedirects or likewise, storing url_path and new_destination or similar, which will be able to map out a history of moving around articles.

Keeping the history would require a global view on the wiki, as moving articles normally
involves as well changing links in other articles. So even a URLPathRedirects table
would not easily allow to restore a state, you still would need to go through all article
revisions and all redirect tables.

What about something simple like:

    +++ src/wiki/models/urlpath.py
    @@ -79,6 +79,8 @@ class URLPath(MPTTModel):
 
         slug = models.SlugField(verbose_name=_('slug'), null=True, blank=True,
                                  max_length=SLUG_MAX_LENGTH)
    +    orig_path = models.CharField(verbose_name=_('original path'), blank=True,
    +                                 max_length=500)
         site = models.ForeignKey(Site)
         parent = TreeForeignKey(
             'self',
    +++ src/wiki/views/article.py
    @@ -491,6 +491,8 @@ class Move(ArticleMixin, FormView):
 
             src_path = self.urlpath.parent.path
             src_slug = self.urlpath.slug + "/"
    +        if (not self.urlpath.orig_path):
    +            self.urlpath.orig_path = self.urlpath.path
             self.urlpath.parent = dest_path
             self.urlpath.slug = form.cleaned_data['slug']
             self.urlpath.save()

An FK would have the disadvantage of not keeping track of moving the parent later on.
And, why new_path? IMHO we want to store the old path for reference.

floemker:
> nschneid commented 12 hours ago
> 
> These sound like nice improvements. Are slugs containing non-ASCII letters still prohibited? This renders them unusable for many languages.

This PR does not change what type of slugs are allowed. It uses the same validation code
for slugs as the article creation code. In the validation code is support for unicode slugs,
but that is not used. Don't know why and if any other code has problems with unicode slugs.

That said, restricting slugs to ASCII makes it IMHO better. Slugs need to be entered worldwide
manually, and when using unicode there that gets much much harder. And additionally unicode
has the problem of identically looking representations for quite different slugs.

benjaoming:
## Suggestion 2 (cont'd)

Thanks for taking time to talk about this @floemker :)

> Keeping the history would require a global view on the wiki, as moving articles normally
involves as well changing links in other articles. So even a URLPathRedirects table
would not easily allow to restore a state, you still would need to go through all article
revisions and all redirect tables.

I think we can relax this part. The critical component is to store the data the right way. Views and user interactions can be made available later (or now if you find it necessary). What concerned me was the data loss indirectly associated with the redirection page.

I agree on the `orig_path`, maybe call it `moved_from_url_path`. But I don't like having it as a CharField because looking up the original URLPath then becomes costly. Because of the way that the move function already works, this means:

 * In case the previous redirecting article is **moved**, it will remain as a new redirect path is created (though only redirecting to one article, but that's okay)
 * In case the text contents of the redirect article is **edited**, data is still available from the moved articles `moved_from_url_path` FK.
 * In case the redirect article is **deleted**, we can do an `on_delete=models.SET_NULL` and basically deleting the redirect article now functions to clean up old redirect fields. In case of the CharField, we wouldn't be able to put constraints like this on it.

I think it's cool to keep the current behaviour all the way, so what I'm suggesting would be to just add an FK `URLPath.moved_from` and fill it in once the redirection Article and its URLPath have been created and are known?

nschneid:
As this PR doesn't affect what slugs are allowed, we don't need to resolve that here (there is a separate issue #576). But to respond:

> Slugs need to be entered worldwide manually, and when using unicode there that gets much much harder. 

If someone wanted to create a wiki entirely in Russian or Chinese, I think they would not want to use a platform that required slugs to be translated or transliterated into ASCII. I'm personally hoping to use the software for a multilingual project in which we definitely need Unicode URLs. In my view it should be a setting on the wiki installation whether non-ASCII slugs are permitted.

> And additionally unicode has the problem of identically looking representations for quite different slugs.

This could partially be solved by normalizing the Unicode string. In #576 there is an example of how to do NFC normalization.

floemker:
> I think it's cool to keep the current behaviour all the way, so what I'm suggesting would be to just add an FK URLPath.moved_from and fill it in once the redirection Article and its URLPath have been created and are known?

That works only if "redirect articles" are created. That's only an option, which is off by default.
Otherwise there is nothing to link to. Even the top moved article can't link to anything,
only the parent is still there at the old location. And because of that, even when linking
always to the old parent the old slug would be lost.

If that is OK, i.e. linking only if "redirect articles" is on, I can do it that way.

benjaoming:
@floemker I understand what you mean. And yes, it would add a lot of complexity to keep storing the old URLPath object for history purposes.. would alter all of our lookups and also require storing a snapshot of the whole parent structure.

> If that is OK, i.e. linking only if "redirect articles" is on, I can do it that way.

Yes, I think that's the only way to do it.

But just one more thing..

Suppose we move 2 articles to 1 location. For instance by merging them or because (as users often do) many people have made articles about the same thing. I know that right now, we can't move an article to a location that already exists, but we *could* potentially add support for this.

That way, we have 2 articles moved to the same location, then the FK should be `URLPath.moved_to = models.ForeignKey('self', null=True, blank=True, related_name='from_set', ...)`.

floemker:
Did now commit an URLPath.moved_from implementation.

floemker:
Did you already decide how to proceed with the open  `str(descendants)`
and with the link questions? Last comments from me concerning that are at

- [floemker#5](https://github.com/floemker/django-wiki/pull/1#issuecomment-303873774) Link question
- [floemker#6](https://github.com/floemker/django-wiki/pull/1#issuecomment-304494271) `str(descendants)` analysis
- [floemker#7](https://github.com/floemker/django-wiki/pull/1#issuecomment-304495431) Possible  `str(descendants)` replacement `descendant.refresh_from_db()`

benjaoming:
@floemker sorry I had this on a TODO for later, but honestly I haven't seen any notifications about the above comments, so thanks for linking!

benjaoming:
Responded in https://github.com/floemker/django-wiki/pull/1

benjaoming:
Passes tests, so merging now and adding release notes.

PR #643, opened 2017-06-06
Global history: Show only last revision of every article

floemker:
Added a new option to the global history plugin to

- show all revisions of all articles or to
- show only the last revision of every article.

Toggling between these two is as well possible on the global history
page via a button.

### DIFF
--- a/src/wiki/plugins/globalhistory/templates/wiki/plugins/globalhistory/globalhistory.html
+++ b/src/wiki/plugins/globalhistory/templates/wiki/plugins/globalhistory/globalhistory.html
@@ -8,11 +8,24 @@
 <h1>{% trans "Global history" %}</h1>
 
 {% spaceless %}
-<p class="lead">
-  {% with paginator.object_list.count as cnt %}
-      List of all <strong>{{ cnt }} changes</strong> in the wiki.
-  {% endwith %}
-</p>
+<div class="row">
+  <div class="lead col-xs-8">
+    {% with paginator.object_list.count as cnt %}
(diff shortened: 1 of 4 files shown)


benjaoming:
Sorry about the direct commit... I wanted to make a PR, but since yesterday's Github outage (just a theory), Github is doing 404s when you choose to create a new branch+PR when editing a file. Is the change okay?

floemker:
> Sorry about the direct commit... I wanted to make a PR, but since yesterday's Github outage (just a theory), Github is doing 404s when you choose to create a new branch+PR when editing a file. Is the change okay?

Yes definitely, looks good.

benjaoming:
I'm curious about the `{% spaceless %}` template block tag that was already there, do you know the purpose?

src/wiki/plugins/globalhistory/templates/wiki/plugins/globalhistory/globalhistory.html
```diff
@@ -8,11 +8,24 @@
 <h1>{% trans "Global history" %}</h1>
 
 {% spaceless %}
```
benjaoming: Why was this already there?

floemker:
> I'm curious about the {% spaceless %} template block tag that was already there, do you know the purpose?

My guess was to make the resulting HTML file smaller.
As I did not see a real disadvantage I left it.

benjaoming:
Yup, it's just not normal to leave indentation and line breaks as it makes the HTML readable and removing it doesn't really save bytes in the bigger picture.

PR #645, opened 2017-06-10
Move article: Two small updates

floemker:
- The testproject database was not migrated from moved_from
  to the new wiki_urlpath.moved_to_id.
- Updated help text on the create form to reflect that changing
  the slug is now possible.


src/wiki/forms.py
```diff
@@ -399,7 +399,7 @@ def __init__(self, request, urlpath_parent, *args, **kwargs):
     slug = WikiSlugField(
         label=_('Slug'),
         help_text=_(
-            "This will be the address where your article can be found. Use only alphanumeric characters and - or _. Note that you cannot change the slug after creating the article."),
+            "This will be the address where your article can be found. Use only alphanumeric characters and - or _.<br>Note: When changing the slug later links pointing to this article are not updated."),
```
benjaoming: Is this text marked safe? Will the `<br>` not be escaped? 

Suggesting to change latter part to: *Note: After changing the slug, links pointing to this article are NOT updated.*

benjaoming:
Thanks for noticing!

src/wiki/forms.py
floemker: @benjaoming 

No, the \<br> is not escaped. It works for me, and the django documentation contains as well

    Field.help_text
    ...
    Note that this value is not HTML-escaped in automatically-generated forms.
    This lets you include HTML in help_text if you so desire. For example:

        help_text="Please use the following format: <em>YYYY-MM-DD</em>."

"After changing the slug" sounds to me as if this is done when creating the new page.
What about:

    Note: If you change the slug later on, links pointing to this article are <b>not</b> updated.

to emphasize even more that this is a problem if the user changes the slug later.
I committed that as well a minute ago.

PR #646, opened 2017-06-10
Use an own paginator

floemker:
The current pagination creates links for every page, which does not work
anymore for many pages. Use instead an own paginator, which creates the
following pagination control:

    < 1 ... 5 6 7 8 9 10 11 12 13 14 ... 10 >

I.e. prev and next links, a link to the first and to the last page, and links
to the current page plus by default four surrounding pages.


src/wiki/core/wiki_paginate.py
```diff
@@ -0,0 +1,30 @@
...
+
+    def __init__(self, *args, **kwargs):
+        """
+        :param side_pages: How many pages should be shown before and after the current page
+        """
+        self.side_pages = kwargs.pop('side_pages', 4)
+        super(WikiPaginator, self).__init__(*args, **kwargs)
+
+    def _get_page(self, *args, **kwargs):
+        self.curPage = super(WikiPaginator, self)._get_page(*args, **kwargs)
```
benjaoming: This assignment contract is too implicit. It means that `page_range` is broken if `_get_page` hasn't been called. Isn't there a way to just call this method directly from `page_range` ?

benjaoming:
Seems necessary and timely, great idea! Made a PR on your branch and added a review comment.

src/wiki/core/wiki_paginate.py
floemker: @benjaoming Calling _get_page() from page_range() unfortunately does not work.
Source from the django paginator (django/core/paginator.py):

    class Paginator(object):
        def _get_page(self, *args, **kwargs):
            """
            Returns an instance of a single page. This hook can be used by
            subclasses to use an alternative to the standard :cls:`Page` object.
            """
            return Page(*args, **kwargs)

    class Page(collections.Sequence):
        def __init__(self, object_list, number, paginator):
            self.object_list = object_list
            self.number = number
            self.paginator = paginator

I.e. the page number is required. And that is exactly what is needed here in the
first place. Based on that I would say that it is guaranteed that _get_page()
is called before page_range().

benjaoming: I tried reordering this a bit 

1. We don't overwrite an weak internal method
1. renamed `curPage` to something with underscores, camelCase isn't used in Python normally :)

It's in https://github.com/floemker/django-wiki/pull/4

floemker: More efficient is always good (int instead of a Page object). Committed.

PR #651, opened 2017-06-25
Since Bootstrap 3 error messages must be shown with alert-danger

floemker:
Since Bootstrap 3 alert-error is renamed to alert-danger. As a result
messages.error() does not use by default the correct class anymore.
This PR changes the settings for the test project, so that that one
uses the alert-danger class.

### DIFF
--- a/testproject/testproject/settings/base.py
+++ b/testproject/testproject/settings/base.py
@@ -13,6 +13,7 @@ from __future__ import unicode_literals
 
 import os
 
+from django.contrib.messages import constants as messages
 from django.core.urlresolvers import reverse_lazy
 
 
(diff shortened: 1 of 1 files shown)


benjaoming:
Oh yeah, that's no good.

But doesn't this interfere with other areas where messages are displayed? For instance in django-admin?

Could also make a template filter that translates?

So that:

```html
          <div class="alert alert-{{ message.tags }}">
            <a class="close" data-dismiss="alert" href="#">&times;</a>
            {{ message }}
          </div>
```

Becomes:

```html
          <div class="alert {{ message.tags|wiki_message_css_class }}">
            <a class="close" data-dismiss="alert" href="#">&times;</a>
            {{ message }}
          </div>
```

benjaoming:
Also, thanks for the PR @floemker !

floemker:
> But doesn't this interfere with other areas where messages are displayed? For instance in django-admin?

@benjaoming Currently not sure. The PR was the result of wondering why my errors where white
and hardly distinguishable from the remaining page and after that searching for the problem.
The first two hits proposed this solution, or a variant of it.

> Also, thanks for the PR @floemker !

Thanks for working continuously on the project!

PR #655, opened 2017-06-26
Add section about setting up account handling 

### LINKED ISSUE #650: 404 on initital user create with /wiki
Python Version: 3.5.2
Django version: 1.10
Wiki version: 0.2.4

Wiki URL's are configured to direct to /wiki as indicated in the installation instructions.

When a user goes to the wiki and creates an account, they are able to successfully, at which point they are directed to the login page. On that initial login, they are sent to http://foo.net/accounts/profile/

This results in a 404 error. Django debug giving:

```
Using the URLconf defined in foo.urls, Django tried these URL patterns, in this order:
^$ [name='index']
^admin/
^wiki/
^notifications/
The current URL, accounts/profile/, didn't match any of these.
```

If they return to http://foo.net/wiki they are logged in and are able to log out/in normally without trouble.

benjaoming:
Fixes #650

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -213,7 +213,7 @@ else:
 # Add any paths that contain custom static files (such as style sheets) here,
 # relative to this directory. They are copied after the builtin static files,
 # so a file named "default.css" will overwrite the builtin "default.css".
-html_static_path = ['_static']
+html_static_path = []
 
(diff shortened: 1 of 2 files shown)


benjaoming:
@Xelinor would you mind giving this a read?

Xelinor:
@benjaoming it looks great and works like a charm.

PR #671, opened 2017-08-06
Prevent processing multiple edit submits in a row.

floemker:
When accidentally double clicking the "Save changes" button in the edit
form all requests are processed, resulting finally in an edit view with

    A new revision of the article was successfully added.

on top and the error

    While you were editing, someone else changed the revision. Your
    contents have been automatically merged with the new contents. Please
    review the text below.

next to it. Prevent this by allowing only one submit action in 3 seconds.

### DIFF
--- a/src/wiki/templates/wiki/edit.html
+++ b/src/wiki/templates/wiki/edit.html
@@ -98,7 +98,13 @@
       $("#article_edit_form").data("changed",true);
     }
     window.onbeforeunload = confirmOnPageExit;
-    $("#article_edit_form").on("submit", function () {
+    var click_time = 0;
+    $("#article_edit_form").on("submit", function (ev) {
+        now = Date.now();
+        elapsed = now-click_time;
+        click_time = now;
+        if (elapsed < 3000)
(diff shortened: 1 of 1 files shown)


floemker:
Something with login / logout when using tox is going wrong.
It looks like logging in works in WebTestCommonMixin

    self.shortcut_login(username=SUPERUSER1_USERNAME,
                                password=SUPERUSER1_PASSWORD)

at least the returned user is admin. But in get_article()

    print("LoggedIn: "+str(request.user.is_authenticated()))

returns False when starting the tests via tox. Starting the tests
via runtests.py directly works fine in contrast. Don't know up to
now what is going on here.

benjaoming:
Alright, sorry about taking such a long time with this - and big thanks for this improvement!

I fully agree that a bit of JavaScript to prevent double-posting is also convenient, and we have the basic server-side implementation in place.

Would be nice at some point also to make this part

> someone else changed the revision.

...sensitive to WHO made the change, "someone else" is in most cases NOT someone else :)

PR #672, opened 2017-08-06
Fix for "Restore attachments"

floemker:
Restoring attachments does not work anymore since
0d84a9cb0093f0669f5a589c0cbb238b19c7a8b1,
as previous_revision is not set anymore for the new
revision when deleting the attachment. But that is needed for
restoring the attachment.  Line from the index.html template:

    if attachment.current_revision.previous_revision.id

The reason seems to be that "created" is used in the pre_save attachment
hook to check if the instance was created or updated, but only the
post_save signal provides a created argument, not prev_save.

So for deciding if the instance is updated or created something else
needs to be used, e.g instance._state.adding.

### DIFF
--- a/src/wiki/plugins/attachments/models.py
+++ b/src/wiki/plugins/attachments/models.py
@@ -197,7 +197,7 @@ def on_revision_delete(instance, *args, **kwargs):
 @disable_signal_for_loaddata
 def on_attachment_revision_pre_save(**kwargs):
     instance = kwargs['instance']
-    if kwargs.get('created', False):
+    if instance._state.adding:
         update_previous_revision = (
             not instance.previous_revision and
             instance.attachment and


floemker:
Seems the test failures are not related. And these failures are the
same as the ones from PR #671:

Something with login / logout when using tox is going wrong.
It looks like logging in works in WebTestCommonMixin

    self.shortcut_login(username=SUPERUSER1_USERNAME,
                                password=SUPERUSER1_PASSWORD)

at least the returned user is admin. But in get_article()

    print("LoggedIn: "+str(request.user.is_authenticated()))

returns False when starting the tests via tox. Starting the tests
via runtests.py directly works fine in contrast. Don't know up to
now what is going on here. Perhaps some dependency change
which broke something?

benjaoming:
This was a rather lengthy investigation :( Turned out that a dependency of another dependency (django-webtest => WebTest) had regressions. See #673

@floemker could you rebase this and other PRs to current master to verify tests?

floemker:
Looks good now. All is green :)

benjaoming:
Alright, cool - btw `git pull --rebase` + `git push -f` is cool for PRs, will avoid merge commits cluttering the history :)

benjaoming:
Thanks for fixing, will add a change log entry

PR #674, opened 2017-08-16
Bump django-functest and django-webtest 

benjaoming:
as proposed by @spookylukey in #673


spookylukey:
Ahem - it might help if I actually *released* django-functest version 1.0.2 ...

spookylukey:
@benjaoming  OK, I released 1.0.2 now, you can retry that TravisCI build now.

PR #677, opened 2017-08-26
URLify: correct comment about which characters are retained

nschneid:


benjaoming:
Thanks for noticing this and making a PR! And sorry about the long response time - I'm normally quicker, but have been around on retreats with some other projects and turned off notifications and forgot to turn them back on :)

Issue #214, opened 2013-11-20
Add alternate display text to attachment


chadmv:
It would be nice to be able to add alternate text for an attachment.  Currently it creates a link using the file name of the attachment.

```
# Currently limited to this:
[attachment:17]

# Propose something like
[An interesting pdf](attachment:17)
# or
[attachment:17 title:"A cool link"]
```

mkane848:
Running into this as well, is there a supported way to either overwrite the default filename or force it to display the Description instead?

benjaoming:
This proposal is better because it doesn't interfere with the existing link syntax.

    [attachment:17 title:"A cool link"]

PR is welcome :) I think it can be implemented entirely within `wiki/plugins/attachments/render.html` and `wiki/plugins/attachments/markdown_extensions.py`. But please add tests!

PR #678, opened 2017-08-27
Extended attachment markdown macro.

floemker:
Extended markdown syntax for attachments:

    [attachment:id title:"text" size]

`title` and `size` are optional. If `title` is given it replaces the default link text, the
file name. If `size` is given the file size of the attachment is appended to the link text.

This is a superset of issue #214.


benjaoming:
Thanks, this is neat! Have added a release note in the docs for 0.3b3.

PR #679, opened 2017-08-27
Avoid overriding global label style

duvholt:


benjaoming:
Cool, must a mistake to have introduce `.wiki-label` without customizing it. Will add a small release note about it, so people know in case they customized CSS and this now takes precedence.

PR #680, opened 2017-08-30
Don't reference South in the README

jamielennox:
I was looking through django-wiki and was surprised to find South
mentioned in the README and so started looking elsewhere. On closer
investigation South has been removed and so should not be mentioned.

Signed-off-by: Jamie Lennox <jamielennox@gmail.com>


benjaoming:
Yes, not only does it not make sense to mention South anymore, but the whole sentence seems dangling in its context - thanks for noticing!

Issue #676, opened 2017-08-23
BUG: Settings Button not working


kaduuuken:
When clicking on the 'Settings' button of an article, it gives the error message:

```
AttributeError at /_settings/
type object 'Settings' has no attribute 'all'
```

Here is the Traceback:

```
Environment:


Request Method: GET
Request URL: http://127.0.0.1:8000/_settings/

Django Version: 1.11.4
Python Version: 3.6.1
Installed Applications:
('django.contrib.admin',
 'django.contrib.auth',
 'django.contrib.contenttypes',
 'django.contrib.sessions',
 'django.contrib.messages',
 'django.contrib.staticfiles',
 'django.contrib.sites',
 'django.contrib.humanize',
 'django.contrib.flatpages',
 'django_nyt',
 'mptt',
 'sekizai',
 'sorl.thumbnail',
 'wiki',
 'wiki.plugins.attachments',
 'wiki.plugins.notifications',
 'wiki.plugins.images',
 'wiki.plugins.macros',
 'allauth',
 'allauth.account',
 'allauth.socialaccount',
 'allauth.socialaccount.providers.slack',
 'slack_chat_connector',
 'activator',
 'raven.contrib.django.raven_compat',
 'anfema_groups')
Installed Middleware:
('raven.contrib.django.middleware.SentryMiddleware',
 'django.contrib.sessions.middleware.SessionMiddleware',
 'django.middleware.common.CommonMiddleware',
 'django.middleware.csrf.CsrfViewMiddleware',
 'django.contrib.auth.middleware.AuthenticationMiddleware',
 'django.contrib.messages.middleware.MessageMiddleware',
 'django.middleware.clickjacking.XFrameOptionsMiddleware',
 'django.contrib.flatpages.middleware.FlatpageFallbackMiddleware')



Traceback:

File "/Users/k.wende/python/wiki/lib/python3.6/site-packages/django/core/handlers/exception.py" in inner
  41.             response = get_response(request)

File "/Users/k.wende/python/wiki/lib/python3.6/site-packages/django/core/handlers/base.py" in _legacy_get_response
  249.             response = self._get_response(request)

File "/Users/k.wende/python/wiki/lib/python3.6/site-packages/django/core/handlers/base.py" in _get_response
  187.                 response = self.process_exception_by_middleware(e, request)

File "/Users/k.wende/python/wiki/lib/python3.6/site-packages/django/core/handlers/base.py" in _get_response
[...]

PR #682, opened 2017-09-03
Fix settings with Django 1.11

floemker:
The notifications plugin uses a ModelChoiceField in it's settings. A
ModelChoiceField needs as first argument a QuerySet or initially None,
not a Model. Since Django 1.11 the provided queryset is already used in
the \_\_init\_\_() function, resulting in an AttributeError when the wrong
type is used here.

This fixed issue #676.

### DIFF
--- a/src/wiki/plugins/notifications/forms.py
+++ b/src/wiki/plugins/notifications/forms.py
@@ -110,7 +110,7 @@ class SubscriptionForm(PluginSettingsFormMixin, forms.Form):
     settings_write_access = False
 
     settings = SettingsModelChoiceField(
-        Settings,
+        None,
         empty_label=None,
         label=_('Settings')
     )


benjaoming:
Thanks for fixing this, will get a patch release out ASAP and report back....

benjaoming:
Also wondering: Why didn't this show up in tests....

tomaszgy:
Hey, guys, thanks for a quick patch! Unfortunately, now when clicking on `Settings` button we're getting another error:
```python
TypeError at /wiki/_settings/
build_attrs() got an unexpected keyword argument 'name'
```

It seems like it's due to a change in `build_attrs()` method arguments in Django after 1.10.4:
https://github.com/django/django/blob/master/django/forms/widgets.py#L243

No `**kwargs` anymore.

benjaoming:
True that... and I think we've fixed this before... and once again, I wonder why tests aren't catching this, they should, let's figure it out...

floemker:
Could you post a complete trace? For me opening the settings works
with Django 1.11.4. And I see only calls to build_attrs_compat()
in the code.

benjaoming:
@floemker - it could be related to enabled plugins? Do you have `notifications`? (sorry, am just in the middle of writing release notes and releasing 0.3b3 on a limited time budget)

floemker:
Yes, notifications are enabled. It's with the nearly default test project.
My only local change is an additional URL pattern for the Django debug toolbar.

benjaoming:
@tomaszgy - we'd need the traceback then over in #687 to make progress - thanks for reporting

tomaszgy:
@benjaoming, just realized that error occured with wiki 0.2.4 and Django 1.11.3 which is not a supported combination. Very sorry for the confusion!

benjaoming:
Alright cool, thanks for letting us know!

Issue #668, opened 2017-07-25
Can't remove URLPath from admin


nicoekkart:
When I try to delete a URLPath via the admin, I get the ValueError error "Cannot call get_ancestors on unsaved URLPath instances".

The url is removed after this error, but this prohibits batch removing of url's, since it instantly stops because of the error..

benjaoming:
Thanks for the report @nicoekkart - Are you going to work on this? Could you post the full traceback?

nicoekkart:
I really don't have any time for it right now. 
Anyways, here is the full traceback:
 ```Environment:


Request Method: POST
Request URL: http://localhost:8000/admin/wiki/urlpath/

Django Version: 1.10.3
Python Version: 3.5.2
Installed Applications:
('django.contrib.auth',
 'django.contrib.contenttypes',
 'django.contrib.sessions',
 'django.contrib.messages',
 'django.contrib.sites',
 'django.contrib.admin',
 'django.contrib.redirects',
 'django.contrib.flatpages',
 'django.contrib.staticfiles',
 'django.contrib.humanize',
 'VTK.apps.user',
 'VTK.apps.tools',
 'VTK.apps.tools.financetools',
 'VTK.apps.tools.financetools.internal',
 'VTK.apps.tools.financetools.external',
 'VTK.apps.activities',
 'VTK.apps.site_main',
 'VTK.apps.members',
 'VTK.apps.links',
 'VTK.apps.about',
 'VTK.apps.sport',
 'VTK.apps.shiften',
 'VTK.apps.common',
 'VTK.apps.tools.praesidiumtools',
 'VTK.apps.tools.praesidiumtools.statistics',
 'VTK.apps.tools.praesidiumtools.mailer',
 'VTK.apps.tools.praesidiumtools.scantool',
 'VTK.apps.tools.praesidiumtools.volunteers',
 'VTK.apps.tools.praesidiumtools.yolofood',
 'VTK.apps.tools.cursustools',
 'VTK.apps.tools.cafetools',
 'VTK.apps.cursus',
 'VTK.apps.mail',
 'VTK.apps.photos',
 'VTK.apps.firnale',
 'VTK.apps.onthaal',
 'VTK.apps.thermilan',
 'VTK.apps.fris',
 'VTK.apps.centralauth',
 'VTK.apps.ads',
 'VTK.apps.elections',
 'VTK.apps.cafe',
 'VTK.apps.survey',
 'VTK.apps.career',
[...]

PR #683, opened 2017-09-04
Don't access a deleted URLPath.

floemker:
django-mptt calls \_\_str\_\_() for an already deleted tree. Prevent in this
case a "Cannot call get_ancestors on unsaved URLPath instances" exception.
This is a work around for an MPTT bug which is needed till PR
django-mptt/django-mptt#591 from MPTT is included in all supported MPTT
versions.

This fixes the exception from issue  #668, but is of course only a work around
for the real bug in MPTT.


benjaoming:
Thanks! This fixes #668, right?

benjaoming:
This looks good, thanks for fixing it so neatly and adding the comment, too!

PR #689, opened 2017-09-08
show logout button if ACCOUNT_HANDLING is not set

jenda1:
if ACCOUNT_HANDLING is not set, the logout button is missing in the user menu now. This change fix it


benjaoming:
Sorry about that, seems that changes in Travis means Python 3.5 isn't available as it used to... will have a look...

jenda1:
perfect, I expected the test fails are unrelated with my change, but I was not able to fix the testcases. thanks

PR #690, opened 2017-09-09
Use Tox-Travis for better support matrix on Travis

benjaoming:
Troubleshooting issue found in #689 Travis tests...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,5 @@
 # Because Python 3.6 is available on GCE
-sudo: required
 dist: trusty
-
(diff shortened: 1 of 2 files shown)


benjaoming:
Not quite optimal yet, because it puts all tox builds for a specific Python version inside the same Tox build... makes it less readable.

For instance, if the code doesn't lint, it would be nice to see that up front, rather than having to guess that it takes place in Python 3.4 and then scroll through 5 different Tox builds...

benjaoming:
Way overdue change, should do this to all the other repos on Github/Travis :)

PR #692, opened 2017-09-11
add input plugin

jenda1:
This change adds new plugin 'inputs'. The plugin allows to create forms in pages. It adds two template tags:

- **[input(-type)? (name)?]** add input field to the page, _type_ is type of the input field. So far the types are implemented:
   - **text:** default type if **_-type_** is missing, input is default text field
   - **textarea:** textarea input field
   - **file/files:** file/files input field
- **[get name]:** show the value of the input field _name_. The name could point to another page.
- **[get-all name]:** show values from all users that fill the input field. It is possible to get values from a page with write access only.

The values are store in the database in a new model _Input_.

This plugin is based on _macros_ plugin.


jenda1:
It looks all the checks fails because lambda in _default_attribute_whitelist. Would be possible to update bleach to v2.0?

benjaoming:
@jenda1 

Thanks for opening a PR

Could you perhaps motivate:

1. What is this plugin useful for?
1. Is it supposed to be in the general distribution as a plugin that's maintained in the main repo?

src/wiki/plugins/inputs/models.py
```diff
@@ -0,0 +1,33 @@
...
+@python_2_unicode_compatible
+class Input(ArticlePlugin):
+    owner = models.ForeignKey(
+        compat.USER_MODEL, verbose_name=_('owner'),
+        null=True, related_name='owned_inputs',
+        help_text=_('The author of the input. The owner always has both read access.'),
+        on_delete=models.SET_NULL)
+
+    key = models.CharField(max_length=28)
+    val = models.TextField()
```
benjaoming: I'm not sure if I would really agree that this is a very structured way to organize data. I added this manifesto to guide what I think this project should do to keep its internal architecture nice and to have deployments and plugin ecosystems have the same over-all goals:

https://github.com/django-wiki/django-wiki/blob/master/CONTRIBUTING.md#manifesto

src/wiki/plugins/inputs/settings.py
```diff
@@ -0,0 +1,15 @@
...
+
+from django.conf import settings as django_settings
+
+SLUG = 'inputs'
+
+INPUTS = getattr(
+    django_settings,
+    'WIKI_PLUGINS_INPUTS',
+    ('text', 'text_inline',
+     'password', 'password_inline',
```
benjaoming: I don't think asking for a password with no secure storage mechanism is a good idea.

src/wiki/plugins/inputs/static/wiki/plugins/inputs/inputs.js
```diff
@@ -0,0 +1,131 @@
+$.ajaxSetup({
```
benjaoming: This is essentially a lot of JS code to make it work.

1. It's hard to know what's going on with no comments
1. It won't work without javascript!
1. There is some kind of template rendering going on, it seems redundant when template rendering is something expected to happen through Django?

jenda1:
ok, I'll start new project with the input plugin. Thanks for your suggestions!

BTW do you know some "plugin only" project I could use as a template? I have never created a django plugin yet.

benjaoming:
Think of the plugin as a complete, self-contained Python package.

Two good links for starting your own Python package.

 * Read about package maintenance: https://packaging.python.org/
 * Use the cookie cutter project as a starter template for all your best-practices: https://pyup.io/repos/github/audreyr/cookiecutter-pypackage/ (I linked a specific favorite template, it contains instructions to get started)

PR #697, opened 2017-09-24
Replaced \d with [0-9] in regexes.

atombrella:
`\d` matches everything on https://www.compart.com/en/unicode/category/Nd

https://code.djangoproject.com/ticket/28628


benjaoming:
Thanks for this! Didn't know about that change, but seems like a good idea!

PR #698, opened 2017-09-29
Fixed secondary search bar not submitting

geoffclapp:
The Bug:

To reproduce: Use the search bar, try to search again in the 'secondary' search bar that appears directly above the results. The button doesn't work.

The secondary search bar button was not submitting in Chrome. Changing the type to 'search' fixes it.

### DIFF
--- a/src/wiki/templates/wiki/search.html
+++ b/src/wiki/templates/wiki/search.html
@@ -14,7 +14,7 @@
     <div class="input-group">
       <input type="search" class="form-control search-query" name="q" value="{{ search_query }}" />
       <span class="input-group-btn">
-        <button class="btn btn-default" type="button">
+        <button class="btn btn-default" type="search">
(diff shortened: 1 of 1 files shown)


src/wiki/templates/wiki/search.html
```diff
@@ -14,7 +14,7 @@ <h1 class="page-header">{% trans "Search results for:" %} {{ search_query }}</h1
     <div class="input-group">
       <input type="search" class="form-control search-query" name="q" value="{{ search_query }}" />
       <span class="input-group-btn">
-        <button class="btn btn-default" type="button">
+        <button class="btn btn-default" type="search">
```
benjaoming: This type should be 'submit', as 'search' doesn't exist.

benjaoming:
Fixed in #699, thanks for spotting!

PR #700, opened 2017-10-01
allow plugins to modify Bleach tag lists

jenda1:
some plugins already needs modify the bleach tags whitelist (plugins/images). This changeset add two options to a plugin class:

- `html_whitelist` - list of http tags to be added to bleach whitelist
- `html_attributes` - dictionary of http attributes to be added to bleach whitelist in the format
   _http tag : [ allowed attributes ]_

### DIFF
--- a/src/wiki/core/plugins/registry.py
+++ b/src/wiki/core/plugins/registry.py
@@ -56,7 +56,6 @@ def register(PluginClass):
             'html_whitelist',
             []))
 
-    # TODO: should we sanitize the list somehow or do we trust plugins to 100%?
     _html_attributes.update(
         getattr(
             PluginClass,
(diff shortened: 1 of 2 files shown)


src/wiki/core/plugins/registry.py
```diff
@@ -48,6 +50,19 @@ def register(PluginClass):
             'markdown_extensions',
             []))
 
+    _html_whitelist.extend(
+        getattr(
+            PluginClass,
+            'html_whitelist',
+            []))
+
+    # TODO: should we sanitize the list somehow or do we trust plugins to 100%?
```
benjaoming: It's not so much about trusting plugins - they are executable code sitting on the server!! :) It's about trusting that user input can't be made malicious.

If a plugin decides to allow everything, well.. so could the user in the settings. So it's the same idea - with great reconfigurability comes great responsibility :)

jenda1: right, I removed the comment

PR #703, opened 2017-10-04
Add context to the 'Summary' string

citxx:
Currently the 'Summary' string in django-wiki collides with the same string in django-admin. It results in django-admin's version being used if `django.contrib.admin` was listed earlier in `INSTALLED_APPS`.

At least Russian translation of django-admin's 'Summary' string is quite bad. It may be the case for other languages too.

I didn't find a clean way to make a fallback to the previous translations. It means that 'Summary' won't be translated until new translation are added in transifex and exported to the django-wiki repository. Please tell me if you know how to do it.

### DIFF
--- a/src/wiki/forms.py
+++ b/src/wiki/forms.py
@@ -16,6 +16,7 @@ from django.forms.widgets import HiddenInput
 from django.utils import timezone
 from django.utils.html import conditional_escape, escape
 from django.utils.safestring import mark_safe
+from django.utils.translation import pgettext_lazy
 from django.utils.translation import ugettext_lazy as _
 from django.utils.translation import ugettext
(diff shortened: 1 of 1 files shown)


benjaoming:
@citxx this is exactly the way to do it!

It's going to have to be done every time that a translation deviates from some other installed application's translations AFAIK.

1. Contextualize ambiguous msgid
1. PR is merged
1. Transifex is updated
1. New contextualized string is translated/approved
1. Compiled new django.mo files go into repo and are released

Not sure if 3-4 are necessary yet.

For reference: https://docs.djangoproject.com/en/1.11/topics/i18n/translation/#contextual-markers

benjaoming:
@citxx I've pushed new sources to Transifex

PR #707, opened 2017-10-21
Cached pip installs in Travis.

atombrella:


benjaoming:
Thanks, that should be nice!

Issue #587, opened 2016-12-25
 New Editor 'simplemde' beta


teury:
From https://simplemde.com/  and https://github.com/onepill/django-simplemde

![Alt text](https://fotos.subefotos.com/e1fd4abdbb5075e6fe455809f5e2caeao.png)

1. 
install django-simplemde
```
pip install django-simplemde
```

2
add 'simplemde' to INSTALLED_APPS.
```

INSTALLED_APPS = (
    # ...
   ` 'simplemde',`
    # ...
)
```

3.
In wiki forms.py

Add `from wiki.simplemde.widgets import SimpleMDEEditor`

In class
```

"CreateRootForm", "EditForm", "CreateForm"
```

Replace: 
```
widget=getEditor().get_widget()`   with  `widget=SimpleMDEEditor() 
```
4. 
 Do the same in admin.py

5. optional 'recommended'
You could set SimpleMDE options in settings.py like this:

```
SIMPLEMDE_OPTIONS = {
    'hideIcons':["guide", "heading"],
    'showIcons':["heading-2"],
    'promptURLs': True,
    'spellChecker': False,
    'placeholder': 'haha',
    'status': False,
    'autosave': {
        'enabled': True
    }
}
```

![Alt text](https://fotos.subefotos.com/364b32f7cf9483f46a1d1cb9d77a0ea8o.png)

benjaoming:
Thanks for the suggestion!

Regarding this:

> Download the simplemde folder. And upload it to the project, in the Wiki folder.

External projects should be implemented as dependencies. Why are you suggesting to copy-paste the code?

Do you have any results of having tested this? Is it all working smoothly? For the plugins (images / attachments), too?

teury:
You can also add an application.

` Pip install django-simplemde
`
And use the

`From simplemde.widgets import SimpleMDEEditor`

Has to work.

(Images / attachments), you would have to configure the plugins so that you can see the images in the same editor.

The images are not inserted in the editor when you press insert. Although in the final result everything is the same as the default editor.


I am doing tests, when I have everything working well, I share the final result.

benjaoming:
Thanks, could be nice with a more modern editor!

prikhi:
@teury Any update? My users are asking for a WYSIWYG editor so I will probably be working on something similar soon.

olymk2:
Any reason this has not been merged in yet ?

benjaoming:
> Any reason this has not been merged in yet ?

[...]

Issue #644, opened 2017-06-07
Pytest & Tox improvements


benjaoming:
- [x] Get rid of runtests.py, too much overhead
- [x] Make it possible to run py.test directly on the current virtualenv
- [x] Tox should accept a positional argument of a specific test to run
- [x] Tox.ini should say `usedevelop = false` because it's dangerous to assume the dev environment is the same as the distributed one.

benjaoming:
@azaghal has totally nailed this issue! Thanks!

PR #709, opened 2017-10-30
Add optional other markdown editor choosen in project config.

olymk2:
#587 putting in this PR for comparison and discussion not ready to merge, needs testing nd style fixes at least and some unit tests.


benjaoming:
Could you explain what Martor is and does?

olymk2:
sure martor is just another markdown editor, it used to be called draceditor its what i actually use on my site I thought i used simplemde but must have moved to martor at some point.

Its has a really nice clean interface to it.

https://github.com/agusmakmun/django-markdown-editor

benjaoming:
@olymk2 ah cool, so it's different from SimpleMDE? And the comparison and discussion you refer to is about which editor to use?

olymk2:
yes its different, well the fact that django wiki allows pluggable editor means comparison / discussion over which on to use is not needed.

You can just choose from the ones provided, i am adding these to test using alternative editors and adjust things as needed.

Failing that this can be a starting point for some one else if i get stuck :)

benjaoming:
@olymk2 cool, let's maybe share some criteria for a future default editor? What's your experience here?

I would ideally like something that...:

1. ...is code oriented - we chose Markdown specifically because it's simple and easy to read. This means that we can show code to normal users.
1. ...does not freak out about custom markdown tags like `[image id:n]`
1. ...provides a simple button interface for common tags (like this Github editor) and has keyboard shortcuts
1. ...has proven its stability and can be kept updated and is not expected to break in deployments a year down the road. MarkItUp is really nice because it hasn't broken a single time!
1. ...does not have a single button that doesn't work or is never needed
1. ...integrates with the remaining layout

olymk2:
My experience is I have played around over the years mainly with markdown or org prefer org because its less fractured and has more useful feature but it a bit niche.

For me consistency is what i want, some of the users of the site are not that techy so require previews and button bars, so that the main thing i am after the current setup works but i keep getting asked how to change things because there is no button bar.

line numbers would also be nice, but mainly consistency across the site and ease of use, the default to me does not matter as long as there are options to change, having simplemde on part of my site then martor in another area does not look great.

I would also like to see consistency between create and edit as currently they behave quite different.

But my requirements are very similar to your list above.

benjaoming:
Thanks for sharing your experience with Markdown! Yeah, the lack of an exhaustive button menu bar is a problem, agreed.

The numbered lines are interesting because:

1. They highlight the syntax of lines
1. Potentially they can be used for annotations (which requires another capability-like interface for the Editor class)

olymk2:
Okay got the image bar working with martor now, so feel free to pull and run the test project switch between the three editor, no reason not to leave the default and let users decide.

martor still has at least a styling issue it button bar is quite long so goes under the sidebar, not actually that keen on the sidebar wonder if it could expand or something it takes up quite a bit of the screen width.

anyway be interested if anyone tries and finds any issues, probably need some test, what would we want to test ? this is frontend and it did not look like we do any selenium style tests, to test the the javascript / editor is integrated and working.

olymk2:
I was going to write some selenium tests for my changes, anyone able to tell me how to run the tests ?

all i get is the error below, using make test py.test and any other method i try am i doing something wrong ?

`E   ImproperlyConfigured: Requested setting DEFAULT_INDEX_TABLESPACE, but settings are not configured. You must either define the environment variable DJANGO_SETTINGS_MODULE or call settings.configure() before accessing settings.`

using these instructions http://django-wiki.readthedocs.io/en/latest/development/testing.html

benjaoming:
@olymk2 it's a problem mentioned in  #644 - you should run for instance `tox -e py35-django111` to test with Python 3.5 and Django 1.11

olymk2:
that works unfortunately the selenium tests fail, although i have chromedriver installed and can fire up python an use it myself. guessing this is something todo with tox ?

any other way to run these tests with out tox ?


```
E       selenium.common.exceptions.WebDriverException: Message: unknown error: Chrome failed to start: exited abnormally
E         (Driver info: chromedriver=2.33.506092 (733a02544d189eeb751fe0d7ddca79a0ee28cce4),platform=Linux 4.13.0-16-lowlatency x86_64)

```

benjaoming:
@olymk2 Is it possibly something missing in the tox environment that's created? Remember that it's created according to the requirements in `tox.ini `

benjaoming:
@olymk2 you could try all this again with the current `master`. Stuff has been fixed so it's easier to run tests locally. See: http://django-wiki.readthedocs.io/en/master/development/index.html#get-started

src/wiki/editors/martor.py
```diff
@@ -16,36 +16,32 @@ def get_admin_widget(self, instance=None):
...
-            'martor/js/martor.min.js',
-        )
+
+        if settings.EDITOR_INCLUDE_JAVASCRIPT is False:
+            m.Media.js = ()
         return m
 
     def get_urls(self):
-        return [url(r'^martor/', include('martor.urls')),]
+        return [url(r'^martor/', include('martor.urls')), ]
```
atombrella: It doesn't seem to be necessary?

benjaoming:
Closing this because of the pro-longed staleness, but thanks for the work on this. The big issue seems to remain: To identify an API for plugins to add buttons and behaviors to these editors, either to the editors themselves, or at least have some kind of JS API where the plugins can add their contents at the current cursor position.

PR #712, opened 2017-11-07
Sort search results, images, and attachments.

floemker:
When providing a list of results these results should be sorted to give
consistent results. Order search results and images by the time they
were changed and attachments by there name.

This fixes the Django 1.11

    UnorderedObjectListWarning: Pagination may yield inconsistent results with an unordered object_list: <class '...'> QuerySet.

warning.


benjaoming:
Thanks for this improvement, makes total sense!

PR #714, opened 2017-11-12
Replace distinct with GROUP BY id

duvholt:
The performance of GROUP BY is much better on PostgreSQL
annotate(Count('id')) is the easiest way to generate GROUP BY id.

Tested on a wiki with about 22 000 revisions the SQL query difference is 33542ms vs 81ms when loading the globalhistory page. 

Preferably this should be tested on other setups too.

### DIFF
--- a/src/wiki/managers.py
+++ b/src/wiki/managers.py
@@ -2,7 +2,7 @@ from __future__ import absolute_import, unicode_literals
 
 import django
 from django.db import models
-from django.db.models import Q
+from django.db.models import Q, Count
 from django.db.models.query import EmptyQuerySet, QuerySet
 from mptt.managers import TreeManager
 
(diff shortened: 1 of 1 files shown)


benjaoming:
Intersting! Do you know about any blogs discussing this or why the `DISTINCT` query would be so slow? Does your database somehow not know that `id` is a unique PK?

duvholt:
I read about it in some stackoverflow post, but now I can't find it anymore. 

Here is the output of explain:

```

duvholt:
I did some more digging and it turns out Postgres by default does `distinct` on all the selected fields, and  because of Django this means every field. 

`distinct` has support for specifying fields, but then the order has to match. 
https://docs.djangoproject.com/en/1.11/ref/models/querysets/#distinct

So by replacing `distinct()` with `order_by('id').distinct('id')` to query is faster again. Still a bit slower than by using group by. Also the query has to be sorted by id instead of modified. 
IMO the group by hack still seems preferable.

benjaoming:
Agreed, could you add a release note about this? Just a one-liner -- just to be super-safe in case it does cause problems.. but will be out as a pre-release, so can be tested before final 0.3.

PR #715, opened 2017-11-12
Support fragments in wiki links

### LINKED ISSUE #701: wiki.plugins.links: links to wiki articles don't respect fragment (#) part of the URL
links plugin cannot find an article, if URL fragment is added to its path.

## Reproduction steps

1. Create an article at path `/article-1` and with the following content:

    ```
    # Some header {: #some-header }
    ```

1. Create another article at path `/article-2` and with the following content:

    ```
    [link to Article 1](wiki:/article-1)

    [link to a header of Article 1](wiki:/article-1#some-header)
    ```

1. Observe the second link generated in the second article

**Observed result:** the link is generated without `href` attribute.
**Expected result:** the link should lead to the first article with the `#some-header` fragment attached to the URL.

## System information

```
$ django-admin.py --version
1.10.5
$ python --version
Python 3.5.2
$ uname -a
Linux citxx-yoga-ubuntu 4.4.0-96-generic #119-Ubuntu SMP Tue Sep 12 14:59:54 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux
```

**P.S.** I have some problems with getting a working django project with the master django-wiki branch. But I can try a bit harder if you want me to test it :) #

floemker:
This fixes #701.

### DIFF
--- a/src/wiki/plugins/links/mdx/djangowikilinks.py
+++ b/src/wiki/plugins/links/mdx/djangowikilinks.py
@@ -57,7 +57,7 @@ class WikiPathExtension(markdown.Extension):
         self.md = md
 
         # append to end of inline patterns
-        WIKI_RE = r'\[(?P<linkTitle>[^\]]+?)\]\(wiki:(?P<wikiTitle>[a-zA-Z0-9\./_-]*?)\)'
+        WIKI_RE = r'\[(?P<linkTitle>[^\]]+?)\]\(wiki:(?P<wikiTitle>[a-zA-Z0-9\./_-]*?)(?P<fragment>#[a-zA-Z0-9\./_-]*?)?\)'
         wikiPathPattern = WikiPath(WIKI_RE, self.config, markdown_instance=md)
         wikiPathPattern.md = md
         md.inlinePatterns.add('djangowikipath', wikiPathPattern, "<reference")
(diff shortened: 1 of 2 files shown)


src/wiki/plugins/links/mdx/djangowikilinks.py
```diff
@@ -57,7 +57,7 @@ def extendMarkdown(self, md, md_globals):
         self.md = md
 
         # append to end of inline patterns
-        WIKI_RE = r'\[(?P<linkTitle>[^\]]+?)\]\(wiki:(?P<wikiTitle>[a-zA-Z0-9\./_-]*?)\)'
+        WIKI_RE = r'\[(?P<linkTitle>[^\]]+?)\]\(wiki:(?P<wikiTitle>[a-zA-Z0-9\./_#-]*?)\)'
```
benjaoming: It would be nicer to use a named group for the fragment, otherwise this is allowed: `wiki:slug#blah#blah`

    r'\[(?P<linkTitle>[^\]]+?)\]\(wiki:(?P<wikiTitle>[a-zA-Z0-9\./_-]*?)(?P<anchor>#[a-zA-Z0-9\./_-]*?)\)'

benjaoming:
Thanks for looking into this @floemker - would be nice to be able to refer to headers, although I think it's a bit outside of the scope of normal users, and header IDs aren't really managed as real data, so if someone changes the header, links referring to it will not scroll to the header anymore.

On the other hand, it's by popular demand, and I don't see the harm in this, since it's not advertised to users.

floemker:
I used now a named group for parsing the fragments.
Additionally the tests for the links plugin are now extended.

src/wiki/plugins/links/mdx/djangowikilinks.py
```diff
@@ -118,8 +118,12 @@ def handleMatch(self, m):
                 path = self.config['base_url'][0] + path_from_link
 
         label = m.group('linkTitle')
+        fragment = m.group('fragment')
+        if not fragment:
+            fragment = ""
+
         a = etree.Element('a')
-        a.set('href', path)
+        a.set('href', path+fragment)
```
benjaoming: Because of some new PEP8 rules, this will break - but will fix it up after + write a release note.

PR #717, opened 2017-11-16
Call super().clean instead of accessing self.cleaned_data directly

atombrella:
Not sure what the motivation for not calling clean, but you're missing out on all of django's built-in validation.


benjaoming:
Oh woops, bad mistake, could cause issues.

benjaoming:
Seeing 100% diff coverage and that tests are fine would indicate that none of these were intentional in leaving out `super()`

PR #721, opened 2017-11-17
Use django.test.TestCase.assertRegexpMatches instead of custom Python…

atombrella:
… 2/3 fallback


benjaoming:
Oh right, I was looking for something like that the other day - thanks, great job! :)

PR #722, opened 2017-11-17
Removed six imports for built-ins.

atombrella:
They're redundant.


benjaoming:
I would be careful, for instance `range` is `xrange` or `range` depending on Python version. There is a reason why they are in `six.moves`

PR #724, opened 2017-11-17
Use x in dict instead of x in list(dict.keys())

atombrella:


atombrella:
Mainly due to speedup, and it's more Pythonic.

benjaoming:
Diff coverage is 100% and tests pass, so should be fine - thanks again :)

PR #725, opened 2017-11-17
assertCountEqual is defined in django.test.TestCase as Python 2/3 fal…

atombrella:
…lback.


benjaoming:
Makes total sense! Thanks for another simplification!

PR #726, opened 2017-11-17
String comparison in wiki.forms.clean_slug

atombrella:
It's slighter faster.


benjaoming:
@atombrella cool, notice that I did a "rebase and merge" here, trying to lower the number of merge commits.

PR #730, opened 2017-11-18
Use more specific asserts instead of the generic assertEqual

atombrella:


benjaoming:
All looks good and passes tests (or rather the tests still work) - thanks! :)

PR #731, opened 2017-11-20
Hierarchical search

### LINKED ISSUE #580: Hierarchical search
Is there a way to search only BELOW the current article?

duvholt:
Fixes #580 

Allows searching below a certain article path. Uses urls like `/test/_search?q=test`.
Otherwise it works just like the normal search. 

I'll take a look at adding some tests for this tomorrow.


src/wiki/templates/wiki/base_site.html
```diff
@@ -39,7 +39,14 @@
...
         <div class="navbar-collapse collapse">
-          <form class="navbar-form navbar-right" id="wiki-search-form" method="GET" action="{% url 'wiki:search' %}">
+          <form class="navbar-form navbar-right" id="wiki-search-form" method="GET" action="{% spaceless %}
+              {% if article or urlpath %}
+                {% url 'wiki:search' article_id=article.id path=urlpath.path %}
+              {% else %}
+                {% url 'wiki:search' %}
+              {% endif %}
+            {% endspaceless %}"
+          >
```
benjaoming: Oh interesting! From a user perspective though, this might get confusing because you will be searching in the current hierarchy but without an option to search the full wiki.

This will also be a behavioral change for anyone used to using this for searching the full wiki.

Could we fix this by editing the placeholder text for the input field? For instance `Search in '{{ article.title }}'` vs `Search full wiki` ?

Also, all of this behavior should maybe be conditioned by whether or not there actually exist any nested articles? Because if not, we might as well default to searching globally?

benjaoming:
Seems like a great idea, and I agree.. if you have already clicked some title, it's typical that you want to search only in the current hierarchy.

src/wiki/views/article.py
```diff
@@ -687,7 +688,17 @@ def dispatch(self, request, *args, **kwargs):
...
-        articles = models.Article.objects.filter(
+        articles = models.Article.objects
+        path = self.kwargs.get('path', None)
+        if path:
+            try:
+                self.urlpath = models.URLPath.get_by_path(path)
+                article_ids = self.urlpath.get_descendants(
+                    include_self=True).values_list('article_id')
+                articles = articles.filter(id__in=article_ids)
+            except (NoRootURL, models.URLPath.DoesNotExist):
```
benjaoming: This should `raise Http404`, no?

duvholt: Yeah, it probably should. I'm not sure how to trigger that from a get_queryset method though.

src/wiki/templates/wiki/base_site.html
duvholt: I agree. I think we should clearly label both the search input and the search view.

Also what about adding a checkbox to the search view to force full search? Reddit does it like this:
image

duvholt: >Also, all of this behavior should maybe be conditioned by whether or not there actually exist any nested articles? Because if not, we might as well default to searching globally?

Right now the search also includes the article you searched from, so I think this would seem odd to the user. It think it makes sense to allow searching for text in just one article, even though it's probably easier to just do a CTRL+F search.

benjaoming: > Right now the search also includes the article you searched from

That's cool, I mean at least not catastrophic in any way..

> Also what about adding a checkbox to the search view to force full search? Reddit does it like this:

To me, it's not mandatory to "expand" the task, I think as long as we make sure the user can easily see where she is searching, it's all fine. But it would be nice with some additional features for the search area, just keep in mind that there's limited space in the topbar, and it has to be responsive.. so I don't know? Perhaps a separate PR?

src/wiki/views/article.py
benjaoming: Oh that's fine, if you raise `Http404` anywhere in a view class, it'll get caught in the middleware. So to raise it here is okay.

duvholt: Oh, neat!

benjaoming:
@duvholt have you had a chance to look at the test failure in this one? :)

duvholt:
@benjaoming should be good now!

Not really sure what I did wrong last time as I tried several things, but now it works at least.

benjaoming:
Awesome work, thanks!!! :)

Will report back when a release is up!

PR #734, opened 2017-11-21
Use super instead of calling method on ParentClass

atombrella:
Also deleted a redundant method


benjaoming:
With the multiple inheritance in these classes, I kind of prefer that the call to a super class is explicit?

atombrella:
Yes, but isn't this simulating that these methods are static, which indeed they aren't in the code-base. I think the codecov-patch is not green because there no tests for those things, and I didn't add any in my patch?

benjaoming:
They are called with `self`, that should make it non-static?

atombrella:
The style looks a bit weird to me, and yes, I did see `self`. I haven't seen in the documentation.

atombrella:
Maybe since there isn't consensus about this change, we should close it.

PR #735, opened 2017-11-22
Removed conversions to lists for generators

atombrella:


atombrella:
It should speed up things a bit.

PR #736, opened 2017-11-22
Replaced str.format with str on article-object

atombrella:


atombrella:
I tried with `timeit` on an empty object, so the changed code should be faster. Not sure how much value it brings to the project, though.

PR #737, opened 2017-11-22
Match Widget.render method signature in custom widgets

atombrella:


benjaoming:
That makes sense! What kind of tool did you use to find this? Was it a manual find? In that case, I tip my hat Sir :)

atombrella:
PyCharm highlights these things. I think there's more.

PR #738, opened 2017-11-22
Don't import absolut_import, it's been optional since Python 2.5

atombrella:
https://www.python.org/dev/peps/pep-0328/ only mentions Python pre 2.7

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-from __future__ import absolute_import, print_function, unicode_literals
+from __future__ import print_function, unicode_literals
 
 import inspect
 import os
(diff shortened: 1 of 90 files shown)


docs/conf.py
```diff
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-from __future__ import absolute_import, print_function, unicode_literals
```
benjaoming: Here you delete `print_function`, too.. it makes it a bit hard to review if this is also deleted, because it requires to look through subsequent lines that tests don't touch

atombrella: I'll add it back.

benjaoming: Still awaiting :)

atombrella: I think Github shows the diff in a strange way. It's indeed there, that's at least what I see with "View".

PR #739, opened 2017-11-22
Use methods from markdown.extensions.toc

atombrella:
This module can be simplified more.

### DIFF
--- a/src/wiki/plugins/macros/mdx/toc.py
+++ b/src/wiki/plugins/macros/mdx/toc.py
@@ -24,22 +24,14 @@ SO WE AN JUST DEPEND ON THAT!
 from __future__ import absolute_import, unicode_literals
 
 import re
-import unicodedata
 
 import markdown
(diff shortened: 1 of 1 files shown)


benjaoming:
I recall that the reason why this module looks as it does, is because changes have been done in those functions -- they aren't supposedly raw copies?

See:

https://github.com/atombrella/django-wiki/commit/d3b0417acf8599625e7a196d1e65b62c42436015

benjaoming:
I don't think this change is desirable, unless you wanna investigate the various versions of the Markdown dependency..

atombrella:
@benjaoming The replaced functions are raw copies. Your comment in the top indicates that this extension could be simplified. We can close it, I don't want to break things.

benjaoming:
@atombrella I updated my comment - no, they don't seem to be modified, it's because they are removed in subsequent Markdown releases.

PR #750, opened 2018-01-01
Allow the HTML tag <sup>.

### LINKED ISSUE #749: markdow error <sup id = "fnref: 1"> duplicate (footnotes)
now

Here is an example with wiki == 0.2b2

floemker:
The by default active footnote extension uses the \<sup> tag for
rendering the footnode. \<sup> needs to be allowed for that.

This fixes #749.


benjaoming:
Oh, bummer -- another one of these. TBH, this bleach stuff has really taken a long time to implement, was a costly move.

benjaoming:
Thanks for fixing!

PR #751, opened 2018-01-01
Fixed alignment of the search button to the search input field.

floemker:
Since the addition of the "Searching in.../Searching whole..." label on the search
page the alignment of the search button compared to the search input field is off,
at least under Firefox. This patch moves the label out of the input-group, so that
the button and the input field are correctly aligned again.

Screen shot before (Firefox 57 under Linux):
image

Screen shot after (Firefox 57 under Linux):
image


benjaoming:
Thanks, looks good, could have been as a `<label>` tag, but thanks for putting in a fix on the very first day of the year, and happy new year :tada: :smiley:

Issue #111, opened 2013-02-15
Make caching more intelligent


benjaoming:
Caching current assumes that there is only ONE version of an article, however depending on the logged in user, there may be several versions. The list of article clear showcases the problem by recursively showing article titles that maybe should be hidden to some users.

To avoid caching per-user, we could group caching into:
- Anonymous user cache
- Group member cache
- Owner cache

All plugins responsible for showing content needs to be alert that there are different versions, and not everything can be assumed. A plugin cannot enter a state dependent on user.id if the user is not the owner.

benjaoming:
Another case of weirdness:

If translations (django.mo files) are updated in subsequent releases, auto-translated parts of articles won't get updated.

There are two hard things in computer science: cache invalidation, naming things, and off-by-one errors.

PR #754, opened 2018-01-16
add user argument to article render

jenda1:
to be able customise render page differently for each user/user's group the user has been added as article's render argument. This change has impact to caching - I add a per-user cache

### DIFF
--- a/src/wiki/models/article.py
+++ b/src/wiki/models/article.py
@@ -200,24 +200,33 @@ class Article(models.Model):
                                           preview=preview_content is not None,
                                           user=user))
 
-    def get_cache_key(self, user=None):
+    def get_cache_key(self):
         lang = translation.get_language()
-        key = "wiki:article:{id:d}:{lang:s}".format(
+        return "wiki:article:{id:d}:{lang:s}".format(
             id=self.current_revision.id if self.current_revision else self.id,
-            lang=lang,
+            lang=lang
         )
 
-        if user:
-            key = "{}:{user!s}".format(key, user=user)
+    def get_cached_content(self, user=None):
+        """Returns cached version of rendered article"""
+        user_key = str(user) if user else ""
+        cache_key = self.get_cache_key()
+        cache_content_key = "{}:{}".format(cache_key, user_key)
 
-        return key
+        cached_items = cache.get(cache_key, list())
+        cached_content = cache.get(cache_content_key) if user_key in cached_items else None
 
-    def clear_cache(self, user=None):
-        """
-        Clear rendered content from cache.
-        If user is None, clear cache for all users.
-        """
-        cache.delete(self.get_cache_key(user))
+        if cached_content is None:
+            cached_content = self.render(user=user)
+            cached_items.append(user_key)
+            cache.set(cache_key, cached_items, settings.CACHE_TIMEOUT)
+            cache.set(cache_content_key, cached_content, settings.CACHE_TIMEOUT)
+        else:
+            cached_content = mark_safe(cached_content)
+        return cached_content
+
+    def clear_cache(self):
+        cache.delete(self.get_cache_key())
 
     def get_absolute_url(self):
         urlpaths = self.urlpath_set.all()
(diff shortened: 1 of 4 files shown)


benjaoming:
Seems related to https://github.com/django-wiki/django-wiki/issues/111

src/wiki/models/article.py
```diff
@@ -206,13 +207,20 @@ def get_cache_key(self):
             lang=lang
         )
 
-    def get_cached_content(self):
+    def get_cached_content(self, user=None):
         """Returns cached version of rendered article"""
+        user_key = str(user) if user else ""
         cache_key = self.get_cache_key()
-        cached_content = cache.get(cache_key)
+        cache_content_key = "{}:{}".format(cache_key, user_key)
```
benjaoming: I agree that the cache key should contain a user component, but should this logic be moved to `self.get_cache_key` and then let the method accept a `user` kwarg?

src/wiki/templatetags/wiki_tags.py
```diff
@@ -52,8 +52,11 @@ def wiki_render(context, article, preview_content=None):
 
     if preview_content:
         content = article.render(preview_content=preview_content)
+    elif article.current_revision:
+        content = article.get_cached_content(context.get('user'))
```
benjaoming: Since `user` is a kwarg, it's best to call it with `user=context.get('user')`

src/wiki/templates/wiki/includes/render.html
```diff
@@ -5,13 +5,7 @@
 {% endaddtoblock %}
 
 <div class="wiki-article">
-  {% if not preview %}
-    {% if article.current_revision %}
-      {{ article.get_cached_content }}
-    {% endif %}
-  {% else %}
-    {{ content|default:"" }}
-  {% endif %}
```
benjaoming: Seems correct that the above should better be replaced with something more understandable.

Did this code also cause a bug?

benjaoming:
Thanks for this important contribution!

I would definitely concede that the previous code was lacking comments -- in case you are freshly recollecting the code that you've written regarding cache keys etc, you would be most welcome with a couple of comments on this :)

src/wiki/models/article.py
jenda1: fixed

src/wiki/templates/wiki/includes/render.html
jenda1: yes, get_cached_content does not get 'user' argument with this code. Also wiki_tags:wiki_render() is probably better place that the template (at least it saves 4 lines of code)

src/wiki/templatetags/wiki_tags.py
jenda1: fixed

benjaoming:
Thanks @jenda1 ! Will have quick final view ASAP!

src/wiki/models/article.py
```diff
@@ -189,33 +189,57 @@ class Meta:
...
+            cached_content = cache.get(cache_content_key)
+            if cached_content is not None:
+                return mark_safe(cached_content)
+
+        cached_content = self.render(user=user)
+        cached_items.append(cache_content_key)
+        cache.set(cache_key, cached_items, settings.CACHE_TIMEOUT)
+        cache.set(cache_content_key, cached_content, settings.CACHE_TIMEOUT)
+
+        return mark_safe(cached_content)
```
benjaoming: How does this scale? 1000 users = 1000 entries for a front page article?

jenda1: if 1000 users open the page before cache timeout, that yes (all anonymous users = 1 user). For now it is resource wasting, but I am working on plugin that add  `[ifgrp/ifusr ...] group/user specific content [endifgrp/usr]` syntax.

BTW I believe 1000 entries in cache are cheaper that a hash table for find a duplicates, but I don't have numbers.

benjaoming: @jenda1 
Thanks for clarifying! :)

I think the cache logic would sit better at the core of the plugin system, so that the plugin designer can mark a specific markdown tag as user-specific. But at least such that cache logic never enters the markdown code itself (user domain!)

PR #755, opened 2018-01-22
Django 2.0 support

rsalmaso:
I'm testing django-wiki, and I need django 2 support :)
It needs https://github.com/benjaoming/django-nyt/pull/55 because I cannot monkey patch as I've done for django-mptt (already django 2 compatible on master) or django-functest (https://github.com/django-functest/django-functest/pull/14)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -24,11 +24,11 @@ def get_path(fname):
 
 
 install_requirements = [
-    "Django>=1.11,<2.0",
+    "Django>=1.11,<2.1",
     "bleach>=1.5,<2",
     "Pillow",
-    "django-nyt>=1.0,<1.1",
-    "django-mptt>=0.8.6,<0.9",
+    "django-nyt>=1.1b1,<1.2",
+    "django-mptt>=0.9,<0.10",
     "django-sekizai>=0.10",
     "sorl-thumbnail>=12,<13",
     "Markdown>=2.6,<2.7",
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ python:
 env:
   - LINT="yes"
   - DJANGO="1.11"
+  - DJANGO="2.0"
 
 matrix:
   exclude:
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -1,0 +1,10 @@
+def hacks():
+    from django.test import testcases
+    if not hasattr(testcases, "WSGIServer"):
+        # django-functest: patch django until is released a new version with
+        # https://github.com/django-functest/django-functest/pull/14 merged
+        from django.core.servers.basehttp import WSGIServer
+        testcases.WSGIServer = WSGIServer
+
+
+hacks()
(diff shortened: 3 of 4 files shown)


benjaoming:
@rsalmaso was thinking about this a while back -- and to make life easier, I would propose that we release 0.3 as a final release now, without Django 2.0 support.

And then bump master to track a new `0.4` series with just Django 1.11 + 2.0 support?

rsalmaso:
And maybe without py27 support?
It would be fine!

benjaoming:
@rsalmaso It's a bit too aggressive to cut 2.7, although I would love to -- the idea behind 1.11 is that it's LTS (so still supported!) and so it helps people to bridge to 2.0... and Python 3 not least!

image

rsalmaso:
@benjaoming ok, in 0.4.x support only django>1.11? it would simplify the patch a bit (remove the is_authenticated/is_anonymous patch) and some urls imports.
But liniting should be ported to python3, because now it fails :)

benjaoming:
> ok, in 0.4.x support only django>1.11?

I would argue `>=1.11` because as you can see, 1.11 LTS is supported until April 2020. Think about it like this: Since you yourself just decided to migrate to 2.0, others are still in that same process :) 1.11 support can help them arrive with fewer bumps! :)

> But liniting should be ported to python3, because now it fails :)

I'll have that done, too!

rsalmaso:
yes, it was `>=1.11` and not `>1.11` (and now corrected :) )

rsalmaso:
Hi, I've split this PR in 4: https://github.com/django-wiki/django-wiki/pull/758 https://github.com/django-wiki/django-wiki/pull/759 https://github.com/django-wiki/django-wiki/pull/760 (which work with django >= 1.8) and this which explicit add django 2.0 support.

benjaoming:
Thanks @rsalmaso ! Is there something left in this PR to merge? I guess we need to bump django-nyt once released, right?

rsalmaso:
Yes, it needs a django-nyt compatible with django 2.0 and at least https://github.com/django-wiki/django-wiki/pull/760 plus probably  some cleanup after the 0.3 release.

setup.py
```diff
@@ -34,12 +34,12 @@ def read_file(fname):
...
 requirements = [
-    "Django>=1.8,<2.0",
+    "Django>=1.8,<2.1",
     "bleach>=1.5,<2",
     "Pillow",
-    "django-nyt>=1.0b1",
+    "django-nyt>1.0b6",
     "six",
-    "django-mptt>=0.8.6,<0.9",
+    "django-mptt>=0.8.6",
```
benjaoming: Removing the upper bounds for django-mptt isn't a good idea. This has historically caused brokenness for deployments.

rsalmaso:
I'm back.
PR wip updated, currently it needs a django-nyt with django 2 support release.
I see that python2 is currently dropped :+1:

benjaoming:
@rsalmaso sorry I thought we were waiting for this to release django-nyt with Django 2.0 support: https://github.com/benjaoming/django-nyt/pull/55

benjaoming:
Hi @rsalmaso !

Just wanted to say that the current django-nyt master has been released to PyPi as 1.1b1 :)

rsalmaso:
PR updated!

PR #758, opened 2018-01-29
Added apps config

rsalmaso:
Added app config for django-wiki and its default plugins and use the django apps machinery.
I've kept an import for existing plugins config in main apps.py.

(pr taken out from https://github.com/django-wiki/django-wiki/pull/755)

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -75,7 +75,7 @@ maintain the order due to database relational constraints:
     'wiki.plugins.attachments.apps.AttachmentsConfig',
     'wiki.plugins.notifications.apps.NotificationsConfig',
     'wiki.plugins.images.apps.ImagesConfig',
-    'wiki.plugins.macros',
+    'wiki.plugins.macros.apps.MacrosConfig',
 
 
 Database
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -74,7 +74,7 @@ maintain the order due to database relational constraints:
     'wiki.apps.WikiConfig',
     'wiki.plugins.attachments',
     'wiki.plugins.notifications.apps.NotificationsConfig',
-    'wiki.plugins.images',
+    'wiki.plugins.images.apps.ImagesConfig',
     'wiki.plugins.macros',
 
 
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -72,7 +72,7 @@ maintain the order due to database relational constraints:
     'sekizai',
     'sorl.thumbnail',
     'wiki.apps.WikiConfig',
-    'wiki.plugins.attachments',
+    'wiki.plugins.attachments.apps.AttachmentsConfig',
     'wiki.plugins.notifications.apps.NotificationsConfig',
     'wiki.plugins.images.apps.ImagesConfig',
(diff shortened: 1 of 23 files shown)


docs/installation.rst
```diff
@@ -71,11 +71,11 @@ maintain the order due to database relational constraints:
...
-    'wiki',
-    'wiki.plugins.attachments',
-    'wiki.plugins.notifications',
-    'wiki.plugins.images',
-    'wiki.plugins.macros',
+    'wiki.apps.WikiConfig',
+    'wiki.plugins.attachments.apps.AttachmentsConfig',
+    'wiki.plugins.notifications.apps.NotificationsConfig',
+    'wiki.plugins.images.apps.ImagesConfig',
+    'wiki.plugins.macros.apps.MacrosConfig',
```
benjaoming: Is there some new consensus in Django documentation that `INSTALLED_APPS` should specify application configuration names for external apps with only 1 default configuration?

src/wiki/conf/settings.py
```diff
@@ -272,7 +273,7 @@
     django_settings,
     'WIKI_SEARCH_VIEW',
     'wiki.views.article.SearchView'
-    if 'wiki.plugins.haystack' not in django_settings.INSTALLED_APPS
+    if not apps.is_installed('wiki.plugins.haystack')
```
benjaoming: neat helper function! 🎉

benjaoming:
Really nice stuff, thanks! Was just wondering if the documentation could be kept more simple, or if it's a new custom to specify the AppConfig names in `INSTALLED_APPS`..?

docs/installation.rst
rsalmaso: I don't know, but I relies on docs to avoid `default_app_config` https://docs.djangoproject.com/en/2.0/ref/applications/#configuring-applications

`New applications should avoid default_app_config. Instead they should require the dotted path to the appropriate AppConfig subclass to be configured explicitly in INSTALLED_APPS.`

rsalmaso:
The dotted app config paths are encouraged by django doc, but it doesn't set an deprecation/removal timeline.

benjaoming:
Regarding the linting fail:

```
src/wiki/apps.py:5:1: F401 'wiki.plugins.attachments.apps.AttachmentsConfig' imported but unused
src/wiki/apps.py:6:1: F401 'wiki.plugins.images.apps.ImagesConfig' imported but unused
src/wiki/apps.py:7:1: F401 'wiki.plugins.notifications.apps.NotificationsConfig as NotifcationsConfig' imported but unused
```

I would suggest to just remove these AppConfigs entirely. Reasoning: 1) Docs never mentioned these AppConfig classes, 2) not all apps were mentioned here, 3) we'll release this with a proper version bump, so it's okay to drop these classes.

rsalmaso:
PR cleaned!

benjaoming:
Thanks! Will merge now!

Have released django-nyt as 1.0 and updated the requiremets for django-wiki such that users will not get some arbitrary version, in case django-nyt 1.1 decides to be backwards incompatible.

PR #760, opened 2018-01-29
Set atomic = False for AlterModelTable migrations

rsalmaso:
Add atomic flag for django >= 1.10, on django < 1.10 it does nothing.

(pr taken out from https://github.com/django-wiki/django-wiki/pull/755)

### DIFF
--- a/src/wiki/plugins/attachments/migrations/0002_auto_20151118_1816.py
+++ b/src/wiki/plugins/attachments/migrations/0002_auto_20151118_1816.py
@@ -6,6 +6,8 @@ from django.db import migrations, models
 
 class Migration(migrations.Migration):
 
+    atomic = False
+
     dependencies = [
         ('wiki_attachments', '0001_initial'),
     ]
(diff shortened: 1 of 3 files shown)


benjaoming:
Am a bit curious here... so this type of migration will fail in a transaction? Will `atomic=False` be set automatically in new Django migrations with table alterations?

rsalmaso:
It is really a django 2.0 enforcement: running without this will fail
`In addition, because of a table alteration limitation of SQLite, it’s prohibited to perform RenameModel and RenameField operations on models or fields referenced by other models in a transaction. In order to allow migrations containing these operations to be applied, you must set the Migration.atomic attribute to False.` (https://docs.djangoproject.com/en/2.0/releases/2.0/#foreign-key-constraints-are-now-enabled-on-sqlite), but as `atomic` flag is added in django 1.10 I think it is ok to add support for it in any case.

And no, makemigrations will not add the atomic flag for a renamed element (why? I don't know).

benjaoming:
Thanks again for your wonderful contributions here, will merge this now and check up on the django-nyt dependency.

PR #762, opened 2018-01-29
releases/0.2.x into master (resolved)

benjaoming:


benjaoming:
@rsalmaso I'm thinking that after this, we can release 0.3 final and branch it off into `releases/0.3.x` to make `master` ready for the 0.4 series with Django 2.0 support.

Sounds good?

rsalmaso:
I'm ok!

PR #767, opened 2018-02-02
Small documentation fix

voglster:
The setting name does not match the expected name (dropped the "File") added this to docs and did not change the setting to preserve backward compatibility.


benjaoming:
Thanks for the quick fix that will be helpful for others @voglster ! True, that should have been named in a more systematic way.

PR #769, opened 2018-02-05
Replaced external six with django.utils.six.

atombrella:
Not sure about this file `requirements_readthedocs.txt`.


benjaoming:
> Not sure about this file requirements_readthedocs.txt.

You can totally remove it from there, too! I'm wondering if it's possible to get rid of it entirely... it was originally to make api-docs run, maybe it will work now, anyways.. will figure it out in a separate branch...

PR #770, opened 2018-02-05
Add missing migrations

benjaoming:
These are auto-generated. I try to trust them, but it seems that some of the changes are in `on_delete` constraints.

PR #772, opened 2018-02-06
Add missing migrations

### LINKED ISSUE #771: Missing migrations
I have found this issue both in master revision and in release 0.3

python manage.py makemigrations --dry-run
Migrations for 'wiki':
  /venv/lib/python3.6/site-packages/wiki/migrations/0003_auto_20180206_1353.py
    - Alter field current_revision on article
    - Alter field current_revision on revisionplugin
    - Alter field parent on urlpath
Migrations for 'wiki_attachments':
  /venv/lib/python3.6/site-packages/wiki/plugins/attachments/migrations/0003_auto_20180206_1353.py
    - Alter field current_revision on attachment

benjaoming:
Re-targetting #770 for a patch release instead.

Fixes #771


benjaoming:
These changes happened in: https://github.com/django-wiki/django-wiki/commit/cd7e4a687e38bc9494522b150b8c02bf3c093e2f

In 3 places, `on_delete` was set to a non-default value, but thankfully enough it was not changed to a non-default value in the initial migration. This was probably on purpose by @rsalmaso 

So adding these 3 migrations should be in fine line with existing deployments.

I like the caution of these three `SET_NULL` constraints, and they also make sense in light of fields already being null-able.

I will try it out:

1. [ ] With a newly spun deployment
1. [ ] On an existing deployment upgraded from having run an older version of `0001_initial`

duvholt:
I think you meant `Fixes #771`? 😃

benjaoming:
> I think you meant Fixes #771? 😃

Ah yes changed -- although strictly speaking, merging this PR will also close it 😃

benjaoming:
I still don't really like this change:

 * Deleting a parent article in a `URLPath` should not turn all the children into new roots.
 * Deleting the current revision should spawn some other kind of constraint, rather than leaving the article without a revision
 * (same for plugins)

benjaoming:
So am actually wondering if the `CASCADE` is better to revert to, without the migrations. That way, the admin system will warn users. `RESTRICT` might be the better choice.

PR #773, opened 2018-02-06
Removed TreeManager.get_query_set fallback.

atombrella:
I had a look at this. It seems this can be safely removed. Please evaluate/double-check; the upgrade notes of `django-mptt` document this.


atombrella:
And I only find `get_query_set` in the test-suite of master of `django-mptt`.

PR #774, opened 2018-02-09
Improvements to translation precision

benjaoming:
@citxx this will break existing translations, so I don't know if you think this is a good idea or not.. I don't have much experience translating, and since I cannot verify existing languages, I don't know if translators are adding bad strings and not objecting like you did :)

### DIFF
--- a/src/wiki/plugins/attachments/templates/wiki/plugins/attachments/delete.html
+++ b/src/wiki/plugins/attachments/templates/wiki/plugins/attachments/delete.html
@@ -9,9 +9,7 @@
 {% if attachment.article == article %}
 <h2>{% trans "Delete" %} "{{ attachment.current_revision.get_filename }}"?</h2>
   <p class="lead">
-    {% blocktrans with attachment.original_filename as filename %}
-    The file may be referenced on other articles. Deleting it means that they will loose their references to this file. The following articles reference this file:
-    {% endblocktrans %}
+    {% trans "The file may be referenced on other articles. Deleting it means that they will loose their references to this file. The following articles reference this file:" %}
   </p>
   <ul>
     {% for a in attachment.articles.all %}
--- a/src/wiki/templates/wiki/includes/breadcrumbs.html
+++ b/src/wiki/templates/wiki/includes/breadcrumbs.html
@@ -45,12 +45,12 @@
         {% if urlpath.parent %}
         <a href="{% url 'wiki:create' path=urlpath.parent.path %}" style="padding: 7px;">
           <span class="fa fa-arrow-right"></span>
-          {% trans "New article next to" %} {{ article.current_revision.title }}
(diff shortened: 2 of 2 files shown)


citxx:
I also don't like the idea of breaking existing translations. In other project I use the approach of falling back to the old string if the translation for the new one doesn't exist yet for the current language. So there is a transition period with two strings present at the same time.

Though I have no idea of how hard or easy would it be to do it in django.

benjaoming:
@citxx that sounds interesting! do you know if it's something that's implemented on the Transifex side normally (i.e. the translation platform)?

citxx:
Honestly, I don't know. All the experience I have with Transifex is translating for django-wiki. In my case we did it in the code, but it was ObjC, not even Python.

benjaoming:
@citxx the immediate challenge I can think of is having the old translation available, because I think that pulling in the translations from Transifex would just render an untranslated message, rather than the old one. Perhaps there is a way to mark an altered message to use translations of the previous version across all languages... but I don't know of this option, although it does seem obvious for some cases.

Another way to do it, would be programmatically by fetching the latest sources of all languages, and somehow fetch the translation from the old version, stuff it in to the downloaded sources, and then upload them back... but I can't really see that happening on a regular basis :)

citxx:
Just to make it more clear what am I talking about. In my case it looked like this:

```objc
// OLD_STRING_ID = "OldStringIDDefinedInResourceFile"
// NEW_STRING_ID = "NewStringIDDefinedInResourceFile"

// Initially
NSString text = getLocalizedString(OLD_STRING_ID)

// Intermediate
NSString text = getLocalizedString(NEW_STRING_ID)
if (text == NEW_STRING_ID) { 
    text = getLocalizedString(OLD_STRING_ID);
}

// When all the translations are in place
NSString text = getLocalizedString(NEW_STRING_ID)
```

We used the fact, that `getLocalizedString` returned the string id (not english text) when there are no translation. Changing the string involved two changes in the code: first to use both strings with a fallback and then to remove the reference to the old string.

Still I don't see an easy way to implement the same in django. Yes, it's possible to start using ids instead of the actual strings and do the similar checks. Probably implement some custom functions and template tags for translations with a fallback and set up Transifex to display english translations, not ids. But it requires a lot of effort AFAICS.

benjaoming:
@citxx yes, this is only possible if you can have a very high level of control of the translation framework, for instance setting up a condition of whether or not something is translated is AFAIK not possible in Django's template language where you mark and translate in the same code: `{% trans "this will be translated to the currently active language, but we don't know what comes out" %}`

Another thing is that it leaves behind legacy code that's not nice in any case.

benjaoming:
Thanks for the example :)

benjaoming:
I'll merge this now, as 0.4 isn't out yet, so there might potentially be plenty of time to get it fixed.

benjaoming:
Merging it after this principle: If not now, then when?

PR #775, opened 2018-02-10
Corrected the super().__init__ call in WikiSlugField.

atombrella:


benjaoming:
Maybe it was by-passed intentionally, but without a proper comment?

atombrella:
Aren't they the same, or am I wrong? The build fails due to Travis/pip issues on Django 1.9.

src/wiki/forms.py
```diff
@@ -55,7 +55,7 @@ def __init__(self, *args, **kwargs):
                 validators.validate_unicode_slug,
                 validate_slug_numbers
             ]
-        super(forms.SlugField, self).__init__(*args, **kwargs)
+        super(WikiSlugField, self).__init__(*args, **kwargs)
```
benjaoming: The old expression by-passed the constructor of the parent, I don't know if this could have been intentional, that's what I was flagging...

atombrella: It's the same. In Python 3, you can just place all these `super(X, self)` with `super()`.

benjaoming: The removed code calls the constructor of the parent of `forms.SlugField` and the new line calls the constructor of the parent of `WikiSlugField` (which is the constructor of `forms.SlugField`)

atombrella: I asked on #python on IRC, and was told the old style is wrong. I'm pretty sure there's no behavioural change.

benjaoming: This means "fetch me the parent of `forms.SlugField` and call its `__init__`

     super(forms.SlugField, self).__init__(*args, **kwargs)

This means "fetch me the parent of `WikiSlugField` and call its `__init__`. The parent of `WikiSlugField` is `forms.SlugField`. So you will call the constructor of that one. But the previous line was specifying the parent of `forms.SlugField`, not `forms.SlugField` itself.

     super(WikiSlugField, self).__init__(*args, **kwargs)

benjaoming: I know we are talking past each others' points here, just wanted to make sure you know what I'm trying to explain... maybe it's different for Python 3, I don't know.

benjaoming: Don't you think they are different, looking at this?

```
python manage.py shell
Python 3.5.2 (default, Nov 17 2016, 17:05:23) 
[GCC 5.4.0 20160609] on linux
Type "help", "copyright", "credits" or "license" for more information.
(InteractiveConsole)
>>> from wiki import forms
>>> s = forms.WikiSlugField()
>>> super(forms.WikiSlugField, s)
<super: <class 'WikiSlugField'>, <WikiSlugField object>>
>>> super(forms.forms.SlugField, s)
<super: <class 'SlugField'>, <WikiSlugField object>>
>>> super(forms.forms.SlugField, s).__init__
<bound method CharField.__init__ of <wiki.forms.WikiSlugField object at 0x7f69092b9c88>>
>>> super(forms.WikiSlugField, s).__init__
<bound method SlugField.__init__ of <wiki.forms.WikiSlugField object at 0x7f69092b9c88>>
>>> 
```

benjaoming: Specifically, `super(forms.forms.SlugField, s).__init__` gives you `CharField.__init__`.

atombrella: It's not different! You can check with this. Each of the constructor in this hierarchy is called.  
```
In [2]: WikiSlugField.__mro__
Out[2]: 
(wiki.forms.WikiSlugField,
 django.forms.fields.SlugField,
 django.forms.fields.CharField,
 django.forms.fields.Field,
 object)
```

atombrella: And there's no difference if it's Python 2 or 3.

benjaoming: >  Each of the constructor in this hierarchy is called.

The removed line actually didn't call `django.forms.fields.SlugField` on purpose.. it was skipped :)

benjaoming: You can try this:

```
class A:

    def __init__(self):
        print("A is constructed")


class B(A):

    def __init__(self):
        print("B is constructed")


class C(B):

    def __init__(self):
        super(B, self).__init__()
        print("C is constructed")


C()
```

It will give you:

```
A is constructed
C is constructed
```

atombrella: There's a difference between this toy example and the `Field`-classes. The `Field`-classes all make an explicit call to the `__init__`-method of the parent.

benjaoming: But this one was skipping `forms.SlugField.__init__`, where would it have been called? :)

    super(forms.SlugField, self).__init__(*args, **kwargs)

benjaoming: What I'm saying is: This PR changes something. If you can state that it doesn't seem to change anything because  `forms.SlugField.__init__` is a no-op, then that's fine.. I'm just flagging that it was probably intentional.

atombrella: That would be `super(forms.WikiSlugField, self)`. I see why it has been simplified in Python 3 to just `super()`. Please see the MRO that I posted above.

benjaoming: Yes exactly, the MRO shows the problem:

```python
(wiki.forms.WikiSlugField,
 django.forms.fields.SlugField,   # = super()
 django.forms.fields.CharField,  # = super(forms.fields.SlugField, self)
 django.forms.fields.Field,
 object)
```

atombrella: Yes. Maybe @rsalmaso would have an opinion about this. It looks wrong.

rsalmaso: I think that `super(forms.SlugField, self)` is the correct one here: it tells to skip SlugField.__init__ call and use `CharField` one.
Judging from django/forms/fields.py code, it done to not override the `default_validators` initialization code in in both `WikiSlugField` and `SlugField` when `allow_unicode == True`.

So we can (in my preference order):
1. keep things as is (`super(forms.SlugField, self).__init__`) which is explicit on what to call
2. change the `WikiSlugField` parent from `SlugField` to `CharField` and use `super().__init__`
3. keep parent as now, but add a `kwargs['validators'] = [validate_slug_numbers]` when appropriate

atombrella: I think changing it to `CharField` would be the correct thing to do.

rsalmaso:
LGTM

benjaoming:
Yup, makes sense now that the inheritance is changed

PR #779, opened 2018-02-13
Simplified wiki.core.utils.object_to_json_response.

atombrella:
Probably more tests around this should be added.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,16 +5,12 @@ language: python
 cache: pip
 
 python:
-  - "2.7"
   - "3.4"
   - "3.5"
   - "3.6"
 
 env:
   - LINT="yes"
-  - DJANGO="1.8"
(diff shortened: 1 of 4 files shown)


atombrella:
Probably more tests around this should be added.

benjaoming:
So the current `master` targets only Django 1.11 and Django 2.0.

So you are welcome to make a change in tox.ini, removing Django 1.8 to allow for this change to happen.

benjaoming:
Ref: http://django-wiki.readthedocs.io/en/master/release_notes.html#release-plan

benjaoming:
Oh you enjoyed that :P

benjaoming:
Alright, so `master` is now officially only Python 3.4+ and Django 1.11+

I guess we can look at the coverage report above and it will identify some of the areas that have gone stale after this change..

image

PR #781, opened 2018-02-14
Improved test dependency handling, removal of wrapper testing script, and fixes for running pytest directly (issue #644)

azaghal:
Most of the changes made in order to satisfy issue #644, with main thing being change of how the dependencies are being handled.

The reasoning in change to test and development requirements handling is to reduce duplication of package listing (so, no need to update in multiple places), to pin the dependent libraries at least to minor version, and to make it a bit more streamlined how test dependencies are installed in development and test environments.


azaghal:
Ok, something looks bonkers with coverage, I'll have to look into it :)

azaghal:
Ok, this is related to removal of `usedevelop = true` from tox.ini.

azaghal:
Ok, fixed the issue with coverage not being run. Now, one thing popped to my mind - I could have `tox` as part of development requirements (this way you don't need to do `pip tox install` in your dev environment). Should I go ahead and make modification for this too? For test environments one would still need to install tox, though (which kinda makes sense).

tox.ini
```diff
@@ -28,29 +28,13 @@ commands =
   # Test that there are no migrations needed -- on Django 1.11, we can
   # use --check and remove the '!' which likely doesn't work on Windows
   sh -c '! testproject/manage.py makemigrations --dry-run --exit --noinput'
-  {toxinidir}/runtests.py --basetemp={envtmpdir} --ds=tests.settings --cov=src/wiki --cov-config .coveragerc {posargs}
+  pytest --basetemp={envtmpdir} --ds=tests.settings --cov={envsitepackagesdir}/wiki --cov-config .coveragerc {posargs}
 
-usedevelop = true
+usedevelop = false
```
benjaoming: This is a very important addition!! Great work!!

setup.py
```diff
@@ -34,6 +34,30 @@ def get_path(fname):
...
+    'pytest-runner',
+]
+
+development_requirements = test_requirements + test_lint_requirements
+
+extras_requirements = {
+    'devel': development_requirements,
+    'test': test_requirements,
+    'testlint': test_lint_requirements,
+}
```
benjaoming: I love this pattern!! Really nice overview of requirements, everything maintained in the same place, and the loosest form of dependencies that affect distribution and testing at the same time.

setup.py
```diff
@@ -34,6 +34,30 @@ def get_path(fname):
...
+    'django-functest>=1.0,<1.1',
+    'mock>=2.0,<2.1',
+    'pytest>=3.4,<3.5',
+    'pytest-django>=3.1,<3.2',
+    'pytest-cov>=2.4,<2.5',
+    'pytest-pythonpath>=0.7,<0.8',
+]
+
+test_lint_requirements = [
+    'flake8>=3.5,<3.6',
```
benjaoming: Agreed! Actually flake8 has been introducing changes that broke linting out of no where... it's best to be explicit about which version of flake8 that linting is supposed to conform to (together with specs in setup.cfg)

PR #787, opened 2018-02-17
Switched to Python 3's super() syntax.

atombrella:
I left the one instance in #775 be.

### DIFF
--- a/src/wiki/views/article.py
+++ b/src/wiki/views/article.py
@@ -36,7 +36,7 @@ class ArticleView(ArticleMixin, TemplateView):
 
     @method_decorator(get_article(can_read=True))
     def dispatch(self, request, article, *args, **kwargs):
-        return super(ArticleView, self).dispatch(request, article, *args, **kwargs)
+        return super().dispatch(request, article, *args, **kwargs)
 
     def get_context_data(self, **kwargs):
         kwargs['selected_tab'] = 'view'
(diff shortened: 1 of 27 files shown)


src/wiki/views/article.py
```diff
@@ -565,7 +564,7 @@ def get_context_data(self, **kwargs):
         if 'form' not in kwargs:
             kwargs['form'] = self.get_form()
         kwargs['purge_form'] = kwargs.pop('form', None)
-        return super(Delete, self).get_context_data(**kwargs)
+        return super().get_context_data(**kwargs)
```
benjaoming: This one wasn't calling `Delete.get_context_data` but the parent's `get_context_data`!

src/wiki/views/article.py
```diff
@@ -574,11 +573,11 @@ class Source(ArticleMixin, TemplateView):
 
     @method_decorator(get_article(can_read=True))
     def dispatch(self, request, article, *args, **kwargs):
-        return super(Source, self).dispatch(request, article, *args, **kwargs)
+        return super().dispatch(request, article, *args, **kwargs)
 
     def get_context_data(self, **kwargs):
         kwargs['selected_tab'] = 'source'
-        return ArticleMixin.get_context_data(self, **kwargs)
+        return super().get_context_data(**kwargs)
```
benjaoming: This one is calling `ArticleMixin.get_context_data` explicitly to avoid calling the `get_context_data` of the wrong parent class.

src/wiki/views/article.py
benjaoming: Can't really tell if it's intentional or not though :)

src/wiki/views/article.py
atombrella: Yes, I'll revert that change.

src/wiki/views/article.py
atombrella: Just to be safe, I'll switch back to the old style with `Delete`. The mismatch between `Derived` and `Parent` as the first argument to `super` is fairly weird to me, though.

PR #790, opened 2018-02-18
Added user.is_{authenticated,anonymous} compatibility

rsalmaso:


benjaoming:
Thanks! :)

PR #792, opened 2018-02-18
Python3 only support

rsalmaso:
Cleanup of python2 code


benjaoming:
Re-running a failed Travis job, seems to be coincidence

CC: @atombrella

PR #793, opened 2018-02-18
Removed django.urls import fallback in wiki.models.

atombrella:


rsalmaso:
@atombrella it seems it's the same in https://github.com/django-wiki/django-wiki/pull/795 (which is extract from https://github.com/django-wiki/django-wiki/pull/755) :)

src/wiki/models/__init__.py
```diff
@@ -99,15 +101,9 @@ def reverse(*args, **kwargs):
...
-    shortcuts.reverse = reverse
-    urls.reverse_lazy = reverse_lazy
-except ImportError:
-    pass
+base.reverse = reverse
+base.reverse_lazy = reverse_lazy
+urls.reverse = reverse
+urls.reverse_lazy = reverse_lazy
+shortcuts.reverse = reverse
+urls.reverse_lazy = reverse_lazy
```
rsalmaso: dup of 107 (took me about 27 days to notice it :smile_cat: )

atombrella: OK. I just very quickly looked at the `ImportError`.

atombrella:
No need for duplicate work. It's in #795 as pointed out.

PR #794, opened 2018-02-18
Remove deprecations

rsalmaso:
Use new MIDDLEWARE setting and urls include (django >= 1.11)


benjaoming:
Thanks! :)

PR #795, opened 2018-02-18
Use django.urls where appropriate

rsalmaso:
Use `django.urls` if possible (django >= 1.11 compatible)

### DIFF
--- a/src/wiki/plugins/attachments/urls.py
+++ b/src/wiki/plugins/attachments/urls.py
@@ -1,4 +1,4 @@
-from django.conf.urls import url
+from wiki.compat import url
 from wiki.plugins.attachments import views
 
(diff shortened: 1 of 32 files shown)


src/wiki/plugins/attachments/urls.py
```diff
@@ -1,4 +1,8 @@
 from django.conf.urls import url
+try:
+    from django.urls import re_path as url
+except ImportError:
+    from django.conf.urls import url
```
benjaoming: Maybe it's best to move this logic to `wiki.core.compat` ?

rsalmaso: Ops, use `wiki.compat`.
Should move?

PR #798, opened 2018-02-19
Don't inherit from object. It's default in Python 3.

atombrella:


benjaoming:
Thanks! Oh wow, this codebase is starting to look fresh!

PR #800, opened 2018-02-19
Removed Django < 1.5 guard for list/QuerySet in SettingsModelForm.save.

atombrella:
The whole notifications plugin does not have any tests. Maybe this should be done in this PR, too?


benjaoming:
> The whole notifications plugin does not have any tests. Maybe this should be done in this PR, too?

Yes, it's in a pretty sad state. In case you don't add tests, it would be appreciated if you can confirm having done a manual check with `testproject` :)

atombrella:
That's also a compromise. I tried to start the tests a little while ago (that's when I noticed the clear notifications bug), but setting up the test data was where I ceased working on this. I can try to make a second attempt.

benjaoming:
Am good with the removed code, seems fine! :) And yes, the notifications plugin really needs tests :)

Issue #796, opened 2018-02-18
Factor out wiki.core.compat


benjaoming:
If necessary, use `wiki.compat` instead

https://github.com/django-wiki/django-wiki/blob/master/src/wiki/core/compat.py

CC: @rsalmaso

rsalmaso:
You want to move wiki.core.compat to wiki.compat?

benjaoming:
@rsalmaso yes, no need for two `compat` modules :)

rsalmaso:
Some cleanup in https://github.com/django-wiki/django-wiki/pull/802

PR #802, opened 2018-02-19
Update compat

rsalmaso:
Fixes https://github.com/django-wiki/django-wiki/issues/796


benjaoming:
Thanks for taking care of that! :)

PR #803, opened 2018-02-19
Used importlib.reload instead of imp.reload in wiki_override_settings.

atombrella:


benjaoming:
Thanks, that's deprecated!

PR #804, opened 2018-02-20
Account handling view tests

atombrella:


tests/core/test_views.py
```diff
@@ -529,6 +538,64 @@ def test_update_profile(self):
...
+class LogoutViewTests(RequireRootArticleMixin, DjangoClientTestBase):
+
+    def test_logout_account_handling(self):
+        response = self.client.get(wiki_settings.LOGOUT_URL)
+        messages = list(response.wsgi_request._messages)
+        user = auth.get_user(self.client)
+        self.assertIs(auth.get_user(self.client).is_authenticated(), False)
+        self.assertIsInstance(user, AnonymousUser)
+        self.assertEqual(len(messages), 1)
+        self.assertEqual(messages[0].message, gettext("You are no longer logged in. Bye bye!"))
```
benjaoming: I think this line can be safely removed -- it might be a personal preference? But I think tests should not block that things change and improve, but should test that things don't break. So the essential test is to see that the user was actually logged out as in `self.assertIsInstance(user, AnonymousUser)`.

tests/core/test_views.py
```diff
@@ -529,6 +538,64 @@ def test_update_profile(self):
...
+        self.assertRedirects(response, reverse('wiki:root'))
+
+    def test_log_in(self):
+        response = self.client.post(
+            wiki_settings.LOGIN_URL,
+            {'username': SUPERUSER1_USERNAME, 'password': SUPERUSER1_PASSWORD}
+        )
+        messages = list(response.wsgi_request._messages)
+        self.assertEqual(len(messages), 1)
+        self.assertEqual(messages[0].message, gettext("You are now logged in! Have fun!"))
```
benjaoming: Also a narrow bit too specific for a test IMO :)

tests/core/test_views.py
atombrella: OK. I can remove it. It took me a little while to figure out the messages thing in testing.

atombrella: I don't agree that these things are unimportant to test. They should remain static, so in essence, it's nice to have such things in the test suite. Anyway, I removed them consistently.

benjaoming: I've tried many times that fixing things become an extra burden, because tests have been written so statically that they break for just about anything. I think this test is really fine, but checking the natural language is always a bit too specific for my own personal taste. It's hard to have a rapid development of a project when everything gets too locked down in test specifications. But this is probably a long discussion :)

PR #805, opened 2018-02-20
Removed Python 2.7 from documentation.

atombrella:


README.rst
```diff
@@ -183,7 +183,7 @@ on your host system.
 -  `django-sekizai <https://github.com/ojii/django-sekizai/>`__
 -  `sorl-thumbnail <https://github.com/mariocesar/sorl-thumbnail>`__
 -  `Pillow (Python Imaging Library) <https://pillow.readthedocs.io/en/latest/installation.html>`__
--  Python>=2.7 or Python>=3.4
```
atombrella: Actually, maybe this should be reverted. The matrix in the top does not list Python versions.

benjaoming: Ah no, it's good to have it written somewhere.. I was wondering if I should specify the Python version in the matrix instead of here.. but can always do that later.. for now, I think the important part is to state that this is Python 3 only.

benjaoming: (was wondering about that because I updated the matrix yesterday)

PR #807, opened 2018-02-20
Move configuration check to django check framework

rsalmaso:
Move all configuration checks from `models/__init__.py` to the Django System check framework.

### DIFF
--- a/src/wiki/checks.py
+++ b/src/wiki/checks.py
@@ -1,0 +1,71 @@
+from django.apps import apps
+from django.core.checks import Error
+from wiki.compat import get_default_engine
+
+
+class Tags:
+    required_installed_apps = "required_installed_apps"
+    obsolete_installed_apps = "obsolete_installed_apps"
+    context_processors = "context_processors"
+
+
+REQUIRED_INSTALLED_APPS = (
+    # module name, package name, error code
+    ('mptt', 'django-mptt', 'E001'),
+    ('sekizai', 'django-sekizai', 'E002'),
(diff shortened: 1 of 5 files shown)


src/wiki/checks.py
```diff
@@ -0,0 +1,70 @@
...
+    return errors
+
+def check_for_context_processors(app_configs, **kwargs):
+    errors = []
+    context_processors = get_default_engine().context_processors
+    for context_processor in REQUIRED_CONTEXT_PROCESSORS:
+        if context_processor[0] not in context_processors:
+            errors.append(
+                Error(
+                    'needs %s in TEMPLATE[\'OPTIONS\'][\'context_processors\']' % context_processor[0],
```
atombrella: If you switch to double-quotes, you don't need escape the single-quotes.

src/wiki/checks.py
```diff
@@ -0,0 +1,70 @@
...
+    ('sekizai.context_processors.sekizai', 'E009'),
+)
+
+
+def check_for_required_installed_apps(app_configs, **kwargs):
+    errors = []
+    for app in REQUIRED_INSTALLED_APPS:
+        if not apps.is_installed(app[0]):
+            errors.append(
+                Error(
```
atombrella: I think this would fit on one line, at least the `Error` part.

rsalmaso:
I was pretty sure to have added the tests...

atombrella:
I looked at the test coverage that's added to the PR. LGTM. It's much cleaner than before.

benjaoming:
Seconded, really nice observation and implementation @rsalmaso !

PR #808, opened 2018-02-20
Increased test coverage for UserCreation and DeleteForm.

atombrella:
This also reverts some auto-pep8 in those two forms.

### DIFF
--- a/src/wiki/forms.py
+++ b/src/wiki/forms.py
@@ -408,8 +408,7 @@ class DeleteForm(forms.Form):
         self.has_children = kwargs.pop('has_children')
         super().__init__(*args, **kwargs)
 
-    confirm = forms.BooleanField(required=False,
-                                 label=_('Yes, I am sure'))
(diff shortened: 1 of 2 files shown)


src/wiki/forms.py
```diff
@@ -625,10 +621,10 @@ def __init__(self, *args, **kwargs):
         self.honeypot_fieldnames = "address", "phone"
         self.honeypot_class = ''.join(
             random.choice(string.ascii_uppercase + string.digits)
-            for __ in range(10))
+            for _ in range(10))
         self.honeypot_jsfunction = 'f' + ''.join(
             random.choice(string.ascii_uppercase + string.digits)
-            for __ in range(10))
+            for _ in range(10))
```
benjaoming: Because of `from django.utils.translation import gettext_lazy as _` this is double underscore

atombrella:
Oh, yes, I didn't think of this. In Python 3, the variable in a list comprehension is not leaked.

benjaoming:
Looking good, thanks for the improvement!

PR #809, opened 2018-02-20
List objects to export in models package

rsalmaso:


benjaoming:
Thanks, that makes the wildcards safe! :)

PR #811, opened 2018-02-20
Isort configuration

rsalmaso:
Move .isort.cfg to config.cfg and expanded to be similar to django one.

### DIFF
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -1,2 +1,0 @@
-[settings]
-skip=docs/conf.py
--- a/setup.cfg
+++ b/setup.cfg
@@ -12,3 +12,6 @@ description-file = README.rst
 
 [aliases]
 test=pytest
+
+[isort]
+skip = docs/conf.py
--- a/setup.cfg
+++ b/setup.cfg
@@ -14,4 +14,11 @@ description-file = README.rst
 test=pytest
 
 [isort]
+combine_as_imports = true
+default_section = THIRDPARTY
+include_trailing_comma = true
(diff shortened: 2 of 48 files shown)


benjaoming:
If you run with the new configuration, does it change stuff in the codebase? Try running it through the pre commit hooks?

rsalmaso:
>If you run with the new configuration, does it change stuff in the codebase?
With this config: yes. see https://gist.github.com/rsalmaso/faf19f430c09836653c5c3a36a4d49a2 for a diff

>Try running it through the pre commit hooks?
I don't know (currently I'm using hg with hg-git), maybe yes. I usually run isort/flake8 manually for reasons.

benjaoming:
> I don't know (currently I'm using hg with hg-git), maybe yes. I usually run isort/flake8 manually for reasons.

That's cool -- the pre-commit hooks are not a policy, the `.travis.yml` linting is the actual policy.

benjaoming:
> With second commit: yes.

This should be part of the PR -- no use leaving it for others to find out :) The changes don't look too bad. Would like to see them pass the tests, otherwise seems good to go.

rsalmaso:
>That's cool -- the pre-commit hooks are not a policy, the .travis.yml linting is the actual policy.

Yes, I run them manually on my machine (they can be really slow sometime) and leave all check when I push.

>This should be part of the PR 

Done

benjaoming:
Diff coverage is not 100% because of the haystack changes, which is fine..

PR #816, opened 2018-02-21
Fix urlize markdown extension

azaghal:
This PR revamps the way the urlize markdown extension is implemented. It is *not* backwards compatible, though. While testing the old implementation, there was a number of issues I ran into (like lots of inline samples not working).

One of the big changes is that the link must be a standalone word (e.g. either at beginning/end of text or surrounded by spaces). This should help avoid some issues.

In general, though, one could see this extension (even the old implementation) as somewhat fragile/too optimistic. For example, if you have an OID 1.2.3.4, it'll get recognised as a link.

On a positive note, the extension has a bit more extensive tests in place now.

### DIFF
--- a/src/wiki/plugins/links/mdx/urlize.py
+++ b/src/wiki/plugins/links/mdx/urlize.py
@@ -4,18 +4,20 @@ import markdown
 
 # Regular expression is meant to match the following pattern:
 #
-# [BEGIN_URL][PROTOCOL]HOST[:PORT][/[PATH]][END_URL]
+# [BEGIN][PROTOCOL]HOST[:PORT][/[PATH]][END]
 #
 # Everything except HOST is meant to be optional, as denoted by square
 # brackets.
 #
 # Patter elements are as follows:
 #
-# BEGIN_URL
-#   Either '<' or '('.
+# BEGIN
+#   String preceding the link. Can be empty, or any string that ends
+#   in whitespace, '(', or '<'.
 #
 # PROTOCOL
-#   One of: 'http://', 'https://', 'ftp://', or 'ftps://'.
+#   Syntax defined in https://tools.ietf.org/html/rfc3986 - for
+#   example: 'http://', 'https://', 'ftp://', or 'ftps://'.
 #
 # HOST
 #   Host can be one of: IPv4 address, IPv6 address in full form, IPv6
@@ -31,8 +33,10 @@ import markdown
 #   Additional PATH, including any GET parameters that should be part
 #   of the URL.
 #
-# END_URL
-#   Either '>' or ')'. Should match with same type as BEGIN_URL.
+# END
+#   String following the link. Can be empty, or any string that ends
+#   in whitespace, ')', or '>'. If ')', then must match with '(' in
+#   BEGIN. If '>', then must match with '<' in BEGIN.
 #
 # It should be noted that there are some inconsitencies with the below
 # regex, mainly that:
@@ -44,20 +48,26 @@ import markdown
 # In order to make the regex easier to handle later on, the following
 # named groups are provided:
 #
-# - begin (character denoting beginning of URL)
+# - begin (string coming before the link, including whitespace or
+#   brackets).
 # - url (entire URL that can be used, for example, as actual link for
 #   href).
 # - protocol (protocol, together with the trailing ://)
 # - host (just the host part)
 # - port (just the port number)
 # - path (path, combined with any additional GET parameters)
-# - end (character denoting end of URL)
+# - end (string coming after the link, including whitespace or
+#   brackets)
 #
 URLIZE_RE = (
-    r'(?P<begin>[\(\<])?(?P<url>'  # begin url group
+    # Links must start at beginning of string, or be preceded with
+    # whitespace, '(', or '<'.
+    r'^(?P<begin>|.*?[\s\(\<])'
+
+    r'(?P<url>'  # begin url group
 
     # Leading protocol specification.
-    r'(?P<protocol>http://|https://|ftp://|ftps://|)'
+    r'(?P<protocol>([A-Z][A-Z0-9+.-]*://|))'
 
     # Host identifier
     r'(?P<host>'  # begin host identifier group
@@ -78,7 +88,10 @@ URLIZE_RE = (
     # Optional trailing slash with path and GET parameters.
     r'(/(?P<path>[^\s\[\(\]\)\<\>]*))?'
 
-    r')(?P<end>[\)\>])?'  # end url group
+    r')'  # end url group
+
+    # Links must stop at end of string, or be followed by a whitespace, ')', or '>'.
+    r'(?P<end>[\s\)\>].*?|)$'
 )
 
 
@@ -92,14 +105,12 @@ class UrlizePattern(markdown.inlinepatterns.Pattern):
         """
 
         # Ensure links are matched only if they stand on their own to avoid bad matches etc.
-        return re.compile(r'^(|.*?\s)%s(\s.*?|)$' % URLIZE_RE, re.DOTALL | re.UNICODE | re.IGNORECASE)
+        return re.compile(URLIZE_RE, re.DOTALL | re.UNICODE | re.IGNORECASE)
 
     def handleMatch(self, m):
         """
         Processes match found within the text.
         """
-
-        matched_string = m.group(0)
 
         protocol = m.group('protocol')
 
@@ -111,12 +122,26 @@ class UrlizePattern(markdown.inlinepatterns.Pattern):
 
         # If opening and ending character for URL are not the same,
         # return text unchanged.
-        if begin_url == '<' and end_url != '>' or begin_url == '(' and end_url != ')':
-            return matched_string
+        if begin_url:
+            begin_delimeter = begin_url[-1]
+        else:
+            begin_delimeter = ''
+        if end_url:
+            end_delimeter = end_url[0]
+        else:
+            end_delimeter = ''
+
+        if (
+                begin_delimeter == '<' and end_delimeter != '>' or
+                begin_delimeter == '(' and end_delimeter != ')' or
+                end_delimeter == ')' and begin_delimeter != '(' or
+                end_delimeter == '>' and begin_delimeter != '<'
+        ):
+            return url
 
         # If no supported protocol is specified, assume plaintext http
         # and add it to the url.
-        if protocol not in ('http://', 'https://', 'ftp://', 'ftps://'):
+        if protocol == '':
             url = 'http://' + url
 
         # Convenience link to distinguish external links more easily.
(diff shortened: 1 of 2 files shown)


src/wiki/plugins/links/mdx/urlize.py
```diff
@@ -2,98 +2,138 @@
...
 
     def handleMatch(self, m):
-        url = m.group(2)
+        """
+        Processes match found within the text.
+        """
+
+        matched_string = m.group(0)
 
-        if url.startswith('<'):
```
azaghal: Actually, this one might not be correctly handled in my new code - what was the intention of this specific check in the original code? I might need to remove special handling for URLs like '<example.com>' or '(example.com)'? Not sure.

benjaoming:
Haven't arrived to reviewing this quite yet, sorry -- might be a couple of days.

> It is not backwards compatible

This part, could you clarify?

benjaoming:
Really exciting and needed improvement btw!

azaghal:
Apropos backwards incompatibilities, well, the following comes to my mind:

- Old code was not able to pass a number of the listed inline samples. It'd produce invalid results, so in a way my code would pass some of those.
- I've removed requirement for matching only `www.DOMAIN` as a FQDN link. Now any kind of FQDN will work.
- I've introduced requirement to have the link standing as standalone string. E.g. it must be at text beginning, text ending, line beginning, or have spaces (any sort of spaces) around it. This helps avoid some issues related to partial matches and the like, for example `-invalid.example.com` would end-up being `<p>-invalid<a href="http://example.com/">example.com</a><p>` in the old code, while with my code it will not end-up being matched at all (can't begin FQDN with a leading dash).
- The matching logic for IPv6 has been changed, since I ran into some kind of loop with the old pattern. Not sure if my new code is the right thing to do. Had a look at some IPv6 regexes out there, but I think they're major PITA to understand and maintain long-term.

I guess one way to check would be to run the new tests against the old code to see what the differences are.

benjaoming:
@azagha Thanks for explaining!l That's really cool, thanks! Am not sure how later versions of Django's `urlize` template filter works, but perhaps it is also worth a look.

Sounds like the backwards incompatibilities are desirable, so I wouldn't call them incompatibilities -- maybe just "improvements" :)

tests/plugins/links/test_urlize.py
```diff
@@ -0,0 +1,231 @@
...
+class TestUrlizeExtension:
+
+    def setup_method(self):
+        self.md = markdown.Markdown(extensions=[UrlizeExtension()])
+
+    @pytest.mark.parametrize("markdown_text, expected_output", FIXTURE_POSITIVE_MATCHES)
+    def test_positive_matches(self, markdown_text, expected_output):
+        assert self.md.convert(markdown_text) == expected_output
+
+    @pytest.mark.parametrize("markdown_text, expected_output", FIXTURE_NEGATIVE_MATCHES)
```
atombrella: Isn't this equivalent/comparable to `unittest.subTest`? I'm not sure how @benjaoming feels about different styles in the test suite, but I think choosing one consistent style is easier to maintain.

tests/plugins/links/test_urlize.py
```diff
@@ -0,0 +1,231 @@
...
+    def test_negative_matches(self, markdown_text, expected_output):
+        assert self.md.convert(markdown_text) == expected_output
```
atombrella: See comment above, use `self.assertEqual` instead? This is what all the other tests use.

benjaoming: Actually, when you are running `pytest`, it will catch these assertion errors and display the problems just as cleanly as `self.assertEqual`. Maybe even better. I don't want to say that it's necessarily better than the other, because it implies fixing up the remaining codebase or having some policy.. I would think both are fine :)

tests/plugins/links/test_urlize.py
benjaoming: Yup, I would agree that if `unittest` can do it, that's option 1.. then what `pytest` can offer would be second choice. But I think it's highly highly imaginative to think that one would want to change from `pytest` in a near future. Seems unlikely that a better test suite pops up or that `pytest` makes breaking API changes.

src/wiki/plugins/links/mdx/urlize.py
```diff
@@ -2,98 +2,138 @@
...
+# - path (path, combined with any additional GET parameters)
+# - end (character denoting end of URL)
+#
+URLIZE_RE = (
+    r'(?P<begin>[\(\<])?(?P<url>'  # begin url group
 
->>> md.convert('go to <http://example.com> now!')
-'<p>go to <a href="http://example.com">http://example.com</a> now!</p>'
+    # Leading protocol specification.
+    r'(?P<protocol>http://|https://|ftp://|ftps://|)'
```
benjaoming: Could also be `onion` :P

Actually this might as well be agnostic to the protocol? https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#Syntax

src/wiki/plugins/links/mdx/urlize.py
```diff
@@ -2,98 +2,138 @@
...
 
->>> md.convert('del.icio.us')
-'<p>del.icio.us</p>'
+    r'[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}|'  # IPv4, match before FQDN
+    r'\[?([A-F0-9]{1,4}:){7}([A-F0-9]{1,4})\]?|'  # IPv6, full form
+    r'\[?:(:[A-F0-9]{1,4}){1,6}\]?|'  # IPv6, leading zeros removed
+    r'([A-F0-9]{1,4}:){1,6}:([A-F0-9]{1,4}){1,6}|'  # IPv6, zeros in middle removed.
+    r'\[?([A-F0-9]{1,4}:){1,6}:\]?|'  # IPv6, trailing zeros removed
+    r'\[?::\]?|'  # IPv6, just "empty" address
+    r'([A-Z0-9]([A-Z0-9-]{0,61}[A-Z0-9])?\.)+([A-Z]{2,6}\.?|[A-Z0-9-]{2,}\.?)|'  # FQDN
```
benjaoming: Great! We retain our `.horse` TLD support :)

src/wiki/plugins/links/mdx/urlize.py
benjaoming: I'm not sure if using `urlize` should cover such. After all, it's meant as a simple shortcut to make certain things clickable. It would be nice if it didn't fire on too many objects and created useless broken links without the user being able to disable it. For all other use cases that it doesn't cover, we can just refer people to the true Markdown syntax: `[text to click](http://whatever.horse)`

src/wiki/plugins/links/mdx/urlize.py
```diff
@@ -2,98 +2,138 @@
...
+    def getCompiledRegExp(self):
+        """
+        Return compiled regular expression for matching the URL
+        patterns. We introduce case-insensitive matching in addition
+        to standard matching flags added by parent class.
+        """
 
-    """ Return a link Element given an autolink (`http://example/com`). """
+        # Ensure links are matched only if they stand on their own to avoid bad matches etc.
+        return re.compile(r'^(|.*?\s)%s(\s.*?|)$' % URLIZE_RE, re.DOTALL | re.UNICODE | re.IGNORECASE)
```
benjaoming: There's already the group called `end` in `URLIZE_RE`, perhaps somehow merge the remaining regular expression logic with that and add a group `start` ?

src/wiki/plugins/links/mdx/urlize.py
azaghal: Ok, I'll update code to not mess with these characters at all.

tests/plugins/links/test_urlize.py
azaghal: I'll have a look at what unittest has to offer in terms of parametrisation. This stuff could be done in loop as well, but thought it was a nice thing to take advantage of functionality.

tests/plugins/links/test_urlize.py
azaghal: If nothing else in the test code uses this, I'll convert this into `unittest` test. Consistency is good. Or, as I like to put it - you want to introduce something new? Then go ahead and do it _everywhere_ :)

tests/plugins/links/test_urlize.py
azaghal: Just had a look at it, that's a nice addition to unittest. I still find pytest a little bit superior, but better to keep everything consistent :)

azaghal: Ok, I take it back a bit. There's a [known missing support for `subTest`](https://github.com/pytest-dev/pytest/issues/1367) in `pytest`.

So, with pytest variant all possible things will be run even if some of them fail, with `pytest + subTest`, tests will stop after first failure.

So... Pick your poison, gentlemen :)

I could "worst case" convert everything to TestCase etc, but still use `pytest`'s parametrisation.

benjaoming: @azaghal then I think it's totally fair to leave it as it is :) I don't want to impose some really rigid way of cutting out obvious benefits of using `pytest` -- after all, it's there to improve testing.

azaghal: Sheesh, ok, another update [from official pytest docs](https://docs.pytest.org/en/latest/unittest.html#pytest-features-in-unittest-testcase-subclasses) - the parametrize thing is not supported.

So... Might need to stick to `pytest`-style unit test unless you're fine with test stopping on first failure?

Again a matter of choose your poison thing, just let me know :)

src/wiki/plugins/links/mdx/urlize.py
azaghal: I actually had to go ahead and verify there _is_ a `.horse` TLD. And congratulations - I'm adding a test for it.

src/wiki/plugins/links/mdx/urlize.py
azaghal: I think begin/end will end-up being dropped once I remove the `()` and `<>` syntax support.

src/wiki/plugins/links/mdx/urlize.py
azaghal: I'll incorporate this change as well.

src/wiki/plugins/links/mdx/urlize.py
azaghal: I ended-up including the `()` and `<>` handling, but in a more robust way.

tests/plugins/links/test_urlize.py
azaghal: I'll be leaving the tests as `pytest`-tailored.

tests/plugins/links/test_urlize.py
azaghal: Went for ease of use instead of consistency. I'll trump over my own words and not do it everywhere, though...

teury:
urlize

Amazing work.

I see a small error that creates links after a point.

benjaoming:
@teury thanks opened an issue for that - are you able to work on it? :)

PR #818, opened 2018-02-22
Use django import_string instead of get_class_from_str

rsalmaso:
Simplify code


benjaoming:
Oh cool, diff coverage 100% :) Thanks for another great modernization @rsalmaso !

PR #822, opened 2018-02-22
Tests cleanup

rsalmaso:


benjaoming:
Makes total sense, thanks a lot for this! :)

PR #823, opened 2018-02-22
Use the full path to the markdown extension with Python's dot notation.

rsalmaso:
Using short names for Markdown's builtin extensions is deprecated.

### DIFF
--- a/src/wiki/conf/settings.py
+++ b/src/wiki/conf/settings.py
@@ -35,15 +35,15 @@ MARKDOWN_SANITIZE_HTML = getattr(
 #:    WIKI_MARKDOWN_KWARGS = {'extension_configs': {'toc': _('Contents of this article')}}
 MARKDOWN_KWARGS = {
     'extensions': [
-        'footnotes',
-        'attr_list',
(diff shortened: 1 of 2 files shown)


src/wiki/conf/settings.py
```diff
@@ -35,15 +35,15 @@
...
+        'markdown.extensions.attr_list',
+        'markdown.extensions.smart_strong',
+        'markdown.extensions.footnotes',
+        'markdown.extensions.attr_list',
+        'markdown.extensions.def_list',
+        'markdown.extensions.tables',
+        'markdown.extensions.abbr',
+        'markdown.extensions.sane_lists',
     ],
     'extension_configs': {
```
benjaoming: This is worth a note in the release notes since it means that if someone added `WIKI_MARKDOWN_KWARGS['extension_configs']`, they might have addressed the modules with just the module name and not the path.

rsalmaso: Added a note

benjaoming: Awesome, thanks :+1: :)

PR #825, opened 2018-02-23
Use template based rendering for custom form widgets

rsalmaso:
Upgrade to new form widget system

### DIFF
--- a/src/wiki/forms.py
+++ b/src/wiki/forms.py
@@ -354,16 +354,16 @@ class SelectWidgetBootstrap(BuildAttrsCompat, forms.Select):
 
 
 class TextInputPrepend(forms.TextInput):
+    template_name = "wiki/forms/text.html"
 
     def __init__(self, *args, **kwargs):
         self.prepend = kwargs.pop('prepend', "")
         super().__init__(*args, **kwargs)
 
-    def render(self, *args, **kwargs):
-        html = super().render(*args, **kwargs)
-        return mark_safe(
-            '<div class="input-group"><span class="input-group-addon">%s</span>%s</div>' %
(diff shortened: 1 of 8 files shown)


atombrella:
This looks really nice! Please keep up all the good work.

benjaoming:
This is really awesome, shows just how much mis-placed code can be moved and reduced through this longly missed feature. I thought about implementing template-rendered form widgets in Django as my university thesis like 5 years ago, but never got around to it :)

benjaoming:
Big thanks for taking the time to do this @rsalmaso !

src/wiki/forms.py
```diff
@@ -285,12 +280,15 @@ def clean(self):
...
 
-class SelectWidgetBootstrap(BuildAttrsCompat, forms.Select):
+class SelectWidgetBootstrap(forms.Select):
     """
     http://twitter.github.com/bootstrap/components.html#buttonDropdowns
     Needs bootstrap and jquery
     """
 
+    template_name = "wiki/forms/select.html"
+    option_template_name = "wiki/forms/select_option.html"
```
benjaoming: I can imagine this must have been confusing with the currently un-designed structure in the `wiki/templates` folder. On one hand, they could live in `wiki/templates/wiki/includes`, and on the other hand, this folder already has form components for the bootstrap template tag sitting in its root, so that would *also* be confusing.

I guess all of this is pending a refactor at some point... but should preferably be announced a bit in advance, since a customized theme can be a real pain to update if we change template structures.

rsalmaso: Well, we can easily refactor until 0.4 "stable" release :joy_cat:

Issue #799, opened 2018-02-19
Deprecate get_pattern


benjaoming:
Same as in `django-nyt`: https://github.com/benjaoming/django-nyt/pull/63

CC: @rsalmaso

rsalmaso:
Done in https://github.com/django-wiki/django-wiki/commit/4026313b0758769568bae7d8030a9db828faae64

benjaoming:
Yay :tada:

PR #827, opened 2018-02-23
Added WikiSite and deprecate wiki.urls.get_pattern

rsalmaso:
Removal of wiki.urls.get_pattern: introduce wiki.site similar to admin site.
It use the same strategy used for django admin default site https://github.com/django/django/commit/da3df5b878d96ef37dcff1377ccbd097b80776d7

### DIFF
--- a/src/wiki/sites.py
+++ b/src/wiki/sites.py
@@ -1,0 +1,183 @@
+from django.apps import apps
+from django.utils.functional import LazyObject
+from django.utils.module_loading import import_string
+from wiki.compat import include, url
+from wiki.core.plugins import registry
+
+
+class WikiSite:
+
+    """
+    Wiki site configurator.
+
+    To customize, you can define your own subclass, either overriding
+    the view providers, or overriding the functions that collect
+    views and set update WikiConfig.default_site to the dotted import path
+    of your customized site.
+    """
+
+    def __init__(self, name='wiki'):
+        from wiki.conf import settings
+        from wiki.views import accounts, article, deleted_list
+
+        self.name = name
+
+        # root view
+        self.root_view = getattr(self, "root_view", article.CreateRootView.as_view())
+        self.root_missing_view = getattr(self, "root_missing_view", article.MissingRootView.as_view())
+
+        # basic views
+        self.article_view = getattr(self, "article_view", article.ArticleView.as_view())
+        self.article_create_view = getattr(self, "article_create_view", article.Create.as_view())
+        self.article_delete_view = getattr(self, "article_delete_view", article.Delete.as_view())
+        self.article_deleted_view = getattr(self, "article_deleted_view", article.Deleted.as_view())
+        self.article_dir_view = getattr(self, "article_dir_view", article.Dir.as_view())
+        self.article_edit_view = getattr(self, "article_edit_view", article.Edit.as_view())
+        self.article_move_view = getattr(self, "article_move_view", article.Move.as_view())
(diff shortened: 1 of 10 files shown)


rsalmaso:
Ok, now is clean.
@atombrella @benjaoming what do you thin?

benjaoming:
Yup, this fixes #799

src/wiki/sites.py
```diff
@@ -0,0 +1,182 @@
...
+        from wiki.views import accounts, article, deleted_list
+
+        self.name = name
+
+        # root view
+        self.root_view_class = article.CreateRootView
+        self.root_missing_view_class = article.MissingRootView
+
+        # basic views
+        self.article_view_class = getattr(self, "article_view_class", article.ArticleView)
```
rsalmaso: I'm thinking about removing the `_class` part and require a view as in
`self.article_view = getattr(self, "article_view", article.ArticleView.as_view())`
Simpler, and less characters.

src/wiki/sites.py
```diff
@@ -0,0 +1,182 @@
...
+        self.article_move_view_class = getattr(self, "article_move_view_class", article.Move)
+        self.article_preview_view_class = getattr(self, "article_preview_view_class", article.Preview)
+        self.article_history_view_class = getattr(self, "article_history_view_class", article.History)
+        self.article_settings_view_class = getattr(self, "article_settings_view_class", article.Settings)
+        self.article_source_view_class = getattr(self, "article_source_view_class", article.Source)
+        self.article_plugin_view_class = getattr(self, "article_plugin_view_class", article.Plugin)
+        self.revision_change_view_class = getattr(self, "revision_change_view_class", article.ChangeRevisionView)
+        self.revision_merge_view_class = getattr(self, "revision_merge_view", article.MergeView)
+
+        self.search_view_class = import_string(getattr(self, "search_view_class", settings.SEARCH_VIEW))
```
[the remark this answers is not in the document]
benjaoming: That's a super idea :+1: That way, specifying views and URLs will be consistent across all functions offered by django-wiki. Also, there are already too many settings.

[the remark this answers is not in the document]
benjaoming: I think you can remove it entirely without a deprecation, but just a release note that it's removed... there are AFAIK not very many haystack users, and there are also indications that it's already broken.

src/wiki/urls.py
```diff
@@ -247,6 +253,11 @@ def get_pattern(app_name="wiki", namespace="wiki", url_config_class=None):
        single Django project.
        https://docs.djangoproject.com/en/dev/topics/http/urls/#topics-http-reversing-url-namespaces
     """
+    import warnings
+    warnings.warn(
+        "wiki.urls.get_pattern is deprecated and will be removed in next version, just include('wiki.urls')",
+        DeprecationWarning
+    )
     if url_config_class is None:
         url_config_classname = getattr(settings, 'URL_CONFIG_CLASS', None)
```
benjaoming: This setting is also deprecated, good idea to state that in `wiki.conf.settings` in the docstring of `URL_CONFIG_CLASS`

src/wiki/sites.py
benjaoming: Yes, since this is an entirely new API, I would think there is no problem in breaking how things were named in the old URL config class?

rsalmaso:
I'll remove SERCH_VIEW in a next PR (both from new and old urls system)

src/wiki/sites.py
rsalmaso: Nobody use them, so there is no BC to preserve :smiley_cat:

PR #828, opened 2018-02-24
Added tests for forms and views in images plugin.

atombrella:


tests/plugins/images/test_forms.py
```diff
@@ -0,0 +1,15 @@
+from django.test import TestCase
+from django.utils.translation import gettext
+
+from wiki.plugins.images.forms import PurgeForm
+
+
+class PurgeFormTest(TestCase):
```
atombrella: I see that there are two forms with more or less the same content, only difference is the label on `confirm`. See `DeleteForm` in the `attachments` plugin. Perhaps it's better to refractor this out to a single form.

benjaoming: Yup, and now that there is test coverage, it would make such a change easier :)

benjaoming:
Also really unique that 58 lines of test code can add >1% points of coverage!!

PR #830, opened 2018-02-24
Use AppConfig.ready to check for sorl.thumbnail in images plugin.

atombrella:
@rsalmaso Perhaps you want to check this, since you replaced a similar pattern in the core.

### DIFF
--- a/src/wiki/plugins/images/apps.py
+++ b/src/wiki/plugins/images/apps.py
@@ -1,8 +1,14 @@
 from django.apps import AppConfig
 from django.utils.translation import gettext_lazy as _
+from django.core.checks import register
+
+from . import checks
 
 
 class ImagesConfig(AppConfig):
     name = 'wiki.plugins.images'
     verbose_name = _("Wiki images")
     label = 'wiki_images'
+
+    def ready(self):
+        register(checks.check_for_required_installed_apps, checks.Tags.required_installed_apps)
--- a/src/wiki/plugins/images/checks.py
+++ b/src/wiki/plugins/images/checks.py
@@ -1,0 +1,13 @@
+from django.apps import apps
+from django.core.checks import Error
+
+
+class Tags:
+    required_installed_apps = "required_installed_apps"
+
+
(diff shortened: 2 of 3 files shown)


src/wiki/plugins/images/apps.py
```diff
@@ -1,8 +1,15 @@
...
 from django.utils.translation import gettext_lazy as _
 
 
 class ImagesConfig(AppConfig):
     name = 'wiki.plugins.images'
     verbose_name = _("Wiki images")
     label = 'wiki_images'
+
+    def ready(self):
+        if not apps.is_installed("sorl.thumbnail"):
```
rsalmaso: This will not use the django check framework.
Better have a checks module which export a function and then register with check framework.
Something as:
```python
rom django.apps import apps
from django.core.checks import Error
class Tags:
  required_installed_apps = "required_installed_apps"
def check_for_required_installed_apps(app_configs, **kwargs):
  errors = []
  if not apps.is_insalled("sorl.thumbnail"):
    errors.append(Error('needs sorl.thumbnail in INSTALLED_APPS', id='wiki_images.E001'))
  return errors
```
and then register them in `ready()`
```python
from django.core.checks import register
from .import checks
def ready(self):
  register(checks.check_for_required_installed_apps, checks.Tags.required_installed_apps)
```

atombrella: Thank you.

src/wiki/plugins/images/checks.py
```diff
@@ -0,0 +1,14 @@
...
+class Tags:
+    required_installed_apps = "required_installed_apps"
+
+
+def check_for_required_installed_apps(app_configs, **kwargs):
+    errors = []
+    if not apps.is_installed("sorl.thumbnail"):
+        errors.append(Error('needs sorl.thumbnail in INSTALLED_APPS', id='wiki_images.E001'))
+    return errors
+
```
rsalmaso: Remove empty line?

PR #836, opened 2018-02-25
Add a default app config

benjaoming:
Was it the intention to leave out this? Because it kind of forces people to specify application configs, although there is a setting (also in Django 2.0) for this: https://docs.djangoproject.com/en/2.0/ref/applications/#for-application-authors

CC: @rsalmaso @atombrella

### DIFF
--- a/src/wiki/__init__.py
+++ b/src/wiki/__init__.py
@@ -17,5 +17,8 @@
 
 from wiki.core.version import get_version
 
+
+default_app_config = 'wiki.apps.WikiConfig'
+
 VERSION = (0, 4, 0, 'alpha', 2)
 __version__ = get_version(VERSION)


benjaoming:
We could/should also add this for the plugins?

In case it isn't specified for `wiki`, then Django will just use the base class `AppConfig` and show: 

```
  File "src/wiki/urls.py", line 10, in <module>
    url(r'^', sites.site.urls),
  File "python3.5/site-packages/django/utils/functional.py", line 238, in inner
    self._setup()
  File "src/wiki/sites.py", line 179, in _setup

AttributeError: 'AppConfig' object has no attribute 'default_site'
```

rsalmaso:
You find what I was missing when working on WikiSite, but couldn't find what :thinking: 
No, it is really needed and left out as an error.

benjaoming:
Cool! I found that you remembered `default_app_config` in all the plugins :)

PR #840, opened 2018-02-28
add conditions plugin

jenda1:
add "conditions" plugin. It adds `[if _condition_]`...something...`[endif]` tags. If _condition_ is evaluated to true, then _something_ is part of wikipage. It allows customise page for different users.  Condition could be based on user name or user's group.

### DIFF
--- a/src/wiki/plugins/conditions/settings.py
+++ b/src/wiki/plugins/conditions/settings.py
@@ -1,0 +1,5 @@
+from __future__ import absolute_import, unicode_literals
+
+# from django.conf import settings as django_settings
+
+SLUG = 'conditions'
--- a/src/wiki/plugins/conditions/mdx/condition.py
+++ b/src/wiki/plugins/conditions/mdx/condition.py
@@ -1,0 +1,62 @@
+import markdown
+import re
+
+class ConditionExtension(markdown.Extension):
+    def extendMarkdown(self, md, md_globals):
+        md.preprocessors.add('dw-conditions', ConditionPreprocessor(md), '>html_block')
+
+
+COND_RE = re.compile(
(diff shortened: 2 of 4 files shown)


src/wiki/plugins/conditions/settings.py
```diff
@@ -0,0 +1,6 @@
+from __future__ import absolute_import, unicode_literals
```
atombrella: Not needed; master is officially Python 3+ only.

benjaoming:
Hi @jenda1 

Thanks for addressing this -- I have a couple of concerns, though:

1. It puts extra and somewhat advanced/special syntax into the Markdown code. I would not encourage this. The mission is to create a Wiki where logic and data isn't embedded in the code, but is left for plugins, which can use UIs and data models to guide the user.
  
   You can read more about this in the "manifesto", which is an attempt to express the direction of the wiki: https://github.com/django-wiki/django-wiki/blob/master/CONTRIBUTING.md

1. If you try to hide something from users, they will always be able to see it inside the source of the wiki article. This would then add extra problems that someone should be able to see an article, but not its source.. this has never been the assumption.

benjaoming:
Btw the Plugin API has never changed. It's quite stable. I would encourage you to go nuts, adding your own special plugins that are rendered through placeholders like `[my_plugin arg1 arg2]`. The plugins can contain logic that renders contents specific to the current user session.

jenda1:
Ok, thanks for your explanation.

- jenda

Dne čt 1. 3. 2018 23:32 uživatel Benjamin Bach <notifications@github.com>
napsal:

benjaoming:
@jenda1 was there anything in the explanation that you disagree with? Or are you okay about maintaining this separately?

Thanks for being understanding, really appreciate the effort to contribute!

jenda1:
sorry for late answer, I missed your replay - I am ok with maintain it
separately.

regards,
- jenda

2018-03-30 23:15 GMT+02:00 Benjamin Bach <notifications@github.com>:

PR #842, opened 2018-02-28
Added tests for the wikilinks plugin.

atombrella:
Please don't merge right away. I'm somewhat confused about the replication of the extension provided by `Markdown`, as it appears it's only due to the `md.article` bit.

There's also an option to configure the `html_class` which seems to be another reason for the duplication.

The shebangs are different places in the code-base, they shouldn't be needed. I removed it from the `wikilinks.py` file.


benjaoming:
Simple and effective! Thanks a bunch!

PR #845, opened 2018-03-01
Removed unneeded parentheses.

atombrella:
Just a small style clean-up.


src/wiki/views/deleted_list.py
```diff
@@ -18,7 +18,7 @@ def get_context_data(self, **kwargs):
         article_list = models.Article.objects.all()
         deleted_articles = []
         for article in article_list:
-            if(article.current_revision.deleted):
```
benjaoming: dafuq

PR #846, opened 2018-03-01
Deleted the local copy of django.get_version

atombrella:
It seems the only difference is 

```
if version is None:
    from wiki import VERSION as version
```


atombrella:
I should learn how to run tests before submitting a PR. Anyway, other packages approach this without the verbatim duplication of `django.version`. I think `django-wiki` should try to avoid this, too.

benjaoming:
> Other packages approach this without the verbatim duplication of django.get_version

Hi @atombrella - do you have a suggestion for this?

The general problem is that if you do `pip install django` and you don't have django in your current environment, already unpacked and installed, it will not be able to import it. That's why it was copied.

atombrella:
@benjaoming I noticed the problem when running the tests, it hadn't occurred to me. Anyway, `wagtail` stripped away a lot of stuff from the `get_version` code; this may also be an option. Other packages mostly use a string. `django-guardian` has both a tuple and a string.

benjaoming:
Here's our over-engineered master piece that occurred by everyone having an opinion :smile:  => https://github.com/learningequality/kolibri/blob/develop/kolibri/utils/version.py

benjaoming:
I don't think anything should be done, maybe copy in any updates from upstream Django -- but the reason that I hard-copied this was to avoid getting something that wasn't well-tested. It's been a nightmare maintaining the above mentioned do-it-together version.py!

atombrella:
OK. Let's close it. I think there are other things where the energy is better spent.

benjaoming:
`get_version` should be a feature of `setuptools` or `distutils`, could be interesting to seek consensus for such a tool in officially endorsed projects.

But as you can see from our homebrew in Kolibri, it's actually a kind of rocket science :)

atombrella:
I think `pkg_resources` already has some implementations of PEP 440 and PEP 386 which provide a `Version`-object that can be used for comparison. `pkg_resources.parse_version('0.4a3')` is usable. Anyway, it's not quite in the format that `django.get_version` wants.

PR #847, opened 2018-03-07
HTML colspan attribute matched number of actual columns.

atombrella:


benjaoming:
I think this is allowed? It just means "span everything, regardless of how many cols" ?

atombrella:
I tried to decipher the specifications, but if it *must* match the number of actual columns, or if this is just good style is unclear to me. https://html.spec.whatwg.org/multipage/tables.html#attr-tdth-colspan Yes, I think 100 as an arbitrarily high number is allowed (it's less than 1000).

benjaoming:
Cool, I will just close this. I think for some (without looking into details) that they *might* have been set high because columns are generated by nested templates, plugin logic etc.

Thanks for the keen eye and for investigating the HTML specs :)

PR #849, opened 2018-03-07
Deleted a redundant initialization in wiki.decorators.get_article.

atombrella:
I think the policy is undoing auto-pep8 is also fine.


benjaoming:
Thanks

PR #850, opened 2018-03-07
Removed a duplicate assert in the test-suite.

atombrella:


benjaoming:
Thanks! It might have been intentional to check that the cache is consistent over multiple calls, but since there's no comment about that, it should not remain :)

PR #855, opened 2018-03-13
 Added some tests for the AttachmentRevision model.

atombrella:
`Field` has both `blank` and `null` set to `False`, so I believe this should be dead.


benjaoming:
Not sure about this one -- mainly because it has always had those set to `False`. So maybe there's a way for it to be `None` ?

atombrella:
OK. Something I missed. Instead, I added some tests for some untested bits and pieces in the `models-py`-file.

PR #856, opened 2018-03-15
Mark editor sidebar headline as safe html

neic:
I'm not completely sure this is safe to do. But I don't se any place this is exposed to an end user.


benjaoming:
You can mark it as safe directly in the plugin class if this is needed, no?

neic:
Yes sure `mark_safe()` works. Brainfart, sorry about that.

benjaoming:
No worries, thanks for caring and opening a PR :)

Issue #702, opened 2017-10-04
Upgrade bleach>=2.1


benjaoming:
There seems to be security issues in the major bleach release, but also there seems to be backwards incompatible stuff.

last-partizan:
Maybe even drop bleach alltogether? and replace with https://github.com/matthiask/html-sanitizer, or something else?

I'm trying to upgrade my project to python3, and bleach uses pinned dependency on html5lib with version incompatible with latest xhtml2pdf.

last-partizan:
Yeah, probably html-sanitizer is a bit overkill.

I looked at issues at bleach, and found this https://github.com/mozilla/bleach/issues/255
They was thinking about replacing html5lib, but looks like it not gone too far.

Updating bleach here would fix my incompatibility, so i'll take a look. What do i need to know before i start?

benjaoming:
@last-partizan 

For the upgrade, it's very important that configuration options aren't changed. I don't think that they will change.

Other than that, coverage is reasonably high.. so I would think it works when the tests say so :)

last-partizan:
Updated bleach, but looks like now they removing tags from strings like `</html>only_this`, instead of escaping.
So, one test fails.

What should we do? I think bleach is correct in this case, and it's better to fix test to expect `only_this`.

benjaoming:
@last-partizan if you open a PR for your changes, you will see results of automated tests etc :)

last-partizan:
Done)

benjaoming:
Fixed in #860 - thanks @last-partizan !!

PR #860, opened 2018-04-10
Bleach upgrade

last-partizan:
Upgraded bleach to latest version and fixed tests (issue #702 )


benjaoming:
Thanks @last-partizan !

One question arising from the test: Does it preserve unintentional tags, let's say a user wants to write a math expression: `x<y>z` (y greater than x and z) ?

last-partizan:
Yes, in this case it would be escaped to `'x&lt;y&gt;z&lt;/y&gt;'`

PR #861, opened 2018-04-10
Pin bleach >=2.1,<2.2

benjaoming:
@last-partizan just to make sure, and trusting in the semantic versioning of the bleach team, we should use the series that we know to work. Is that okay for your purpose?

Also added a release note. Can release immediate as another 0.4 version...


last-partizan:
yeah, that's fine!

PR #865, opened 2018-05-10
project fails to start when User model that doesn not have a 'username' field.

trevorpeacock:
In a project with a custom User model that does not have a 'username' field, such as
```
class User(AbstractBaseUser, MyPermissionsMixin):
    email = models.EmailField(_('email address'), unique=True)
    USERNAME_FIELD = 'email'
```
 the project fails to load with
```
 File ".../wiki/views/accounts.py", line 21, in <module>
    from wiki import forms
  File ".../wiki/forms.py", line 571, in <module>
    class UserCreationForm(UserCreationForm):
  File ".../django/forms/models.py", line 277, in __new__
    raise FieldError(message)
django.core.exceptions.FieldError: Unknown field(s) (username) specified for User
```

There are two problems.
1) django-wiki in `forms.UserCreationForm` specifies `UserCreationForm.Meta.fields = ('username', "email")
2) django-wiki ignores a potential custom `UserCreationForm` (as per [reccomendations](https://docs.djangoproject.com/en/1.11/topics/auth/customizing/#custom-users-and-the-built-in-auth-forms). Though it seems Django may not provide a good way of handling this)

A solution is to refer to the custom models `USERNAME_FIELD` attribute (`django.contrib.auth.get_user_model().USERNAME_FIELD`). This appears to work even if the `USERNAME_FIELD` is `email`. Though it may not work on a model where there is no `email` field.

A more general solution may be to inherit from the custom `UserCreationForm`. Getting hold of that is a little messier.
```
from django.contrib import admin
from django.contrib.auth import get_user_model
UserCreationForm = admin.site._registry[get_user_model()].add_form
```
Its unclear to me how robust this solution is.

The submitted pull request combines both options, though only one may be required or appropriate.


benjaoming:
Hi @trevorpeacock - thanks for the contribution.

Just wanted to let you know that in general, these registration and authentication pages are just meant for people to get quickly started.

If you are building your own logic with custom models etc, it's better to disable these views entirely. See: http://django-wiki.readthedocs.io/en/master/settings.html

If the errors are happening, despite that the forms/views aren't in use, that's a serious bug.

src/wiki/forms.py
```diff
@@ -567,7 +568,7 @@ class SearchForm(forms.Form):
         required=False)
 
 
-class UserCreationForm(UserCreationForm):
+class UserCreationForm(admin.site._registry[User].add_form if User in admin.site._registry else UserCreationForm):
```
benjaoming: Is `admin.site._registry` supposed to be accessed externally?

And using the form from the admin doesn't seem like anything we should intend for public user registration.

trevorpeacock:
The exception occurs on initialisation, not on a request/view. It occurs with `settings.WIKI_ACCOUNT_HANDLING = False`.

I agree, the `admin.site._registry[get_user_model()].add_form` solution is
1) ugly
2) does access an the internal _registry attribute, making the solution potentially brittle
3) makes django-wiki dependent on the good behavior of custom user model definitions.
4) will not work if the model is not added to the admin.site, or added to a non-default admin.site

While I don't understand the exact mechanism, it seems Django does some pre-checks for any forms loaded, in particular, checking the fields added to a ModelForm exist in the related Model.

Solutions may be
1) look up fields from User Model. A variation and expansion on the following (This may require re-thinking the email field, and its requirements in django-wiki)
```
class UserCreationForm(UserCreationForm):
    class Meta:
        fields = tuple([User.USERNAME_FIELD,] + User.REQUIRED_FIELDS)
```

2) use `django.contrib.auth.models.User` instead of `django.contrib.auth.get_user_model()`. If the registration/authentication are not intended to be robust solutions where custom user models are used, particularly if they require any particular field names, perhaps no attempt should be made to cater for them.

I have a demonstration project thats the simplest I could come up with, with a (mostly) working custom User, UserManager, and ModelAdmin.
https://github.com/trevorpeacock/django-wiki-customuser-test

trevorpeacock:
So, Proposed solution (I'll be able to spend a little time on this in two weeks)

  1) wrap the `from wiki import forms` in `accounts.py` in an `if settings.WIKI_ACCOUNT_HANDLING:` statement, re-working whatever is required to make that feasible (splitting classes out from forms/accounts to separate files if necessary). Is there a better way than wrapping an import in an if?

  2) either remove `get_user_model()` and build against the default django user class, and refactor as needed

  OR

  2) do what we can to make forms.py a little more robust (probably allowing different username field names, but probably requiring an email address field?). Allowing best of both worlds. "batteries included" works as long as you're not doing anything too out of the ordinary, but you can turn it all off and build yourself if required.

Thoughts?

I've not looked at accounts.py, not sure how big a job it is. Until I do look at it, am I being naive as to how simple a solution this is?

benjaoming:
Hi @trevorpeacock !

Thanks for sharing great ideas! I call them 1a and 1b as two combined alternatives then :)

1a: option one + first 2)
1b: option one + second 2)

1a: This is great because it's strict and doesn't promise something that "might" work. No one can really use "might works" for a lot. And people who are sub-classing `django.contrib.auth.User` are probably also doing their own login system.

1b: it retains the guarantee that batteries are included for `django.contrib.auth` and can be switched off if they don't work - but also doesn't rule out that someone may have sub-classed `User` with compatible changes.

1b is the least intrusive for now. It would simply provide a way to entirely switch off the current system.

I think 1b is easier to implement both for us and for people using django-wiki, so perhaps its the best place to start. Afterwards, it would be possible to "add" 1a on top, limiting the scope of the authentication forms/views to only apply to vanilla `django.contrib.auth`.

benjaoming:
Hi @trevorpeacock - sorry if the feedback got lengthy. Would you be able to work on this?

benjaoming:
Hi!

Thanks for adding more commits to the PR, it seems that it is well on it's way now?

Could you do this from your command line in order to wrap everything into a single commit and lint the code?

```
# Add main repo as "upstream" remote
git remote add upstream https://github.com/django-wiki/django-wiki.git
# Fetch changes
git fetch upstream
# Reset (soft), rolling back all commits and then adding them on top
git reset upstream/master

# Install the pre-commit framework that will lint your code
pip install pre-commit
pre-commit install

# Add everything in to a single commit (provided you don't have local files that shouldn't go in)
git add .

# Now try to commit and see the commit hooks being run
git commit
```

trevorpeacock:
decided its more understandable to split classes dependent on checks to a separate file

trevorpeacock:
I've addressed the concerns @benjaoming raised. I've updated the pull request, though I'm not sure I've done it correctly. All the checks are passing. Can anyone help me progress this issue?

benjaoming:
Sorry that this has been sitting for so long. I have missed out on notifications. Thanks for getting back to it, @trevorpeacock ! Will review again ASAP!

src/wiki/forms.py
```diff
@@ -21,6 +18,10 @@
 from wiki.core.plugins.base import PluginSettingsFormMixin
 from wiki.editors import getEditor
 
+from .forms_account_handling import (  # noqa: ignore=F401 importing additional forms here as a convenience for other modules expecting them here
```
benjaoming: I'm not sure I understand the necessity of this linting bypass?

src/wiki/forms_account_handling.py
```diff
@@ -0,0 +1,89 @@
...
+
+def check_email_field(user_model=User):
+    return isinstance(_get_field(user_model, user_model.get_email_field_name()), EmailField)
+
+
+# django parses the ModelForm (and Meta classes) on class creation, which fails with custom models without expected fields.
+# We need to check this here, because if this module can't load then system checks can't run.
+CustomUser = User \
+    if (settings.ACCOUNT_HANDLING and check_user_field() and check_email_field()) \
+    else django.contrib.auth.models.User
```
benjaoming: Is it a good idea that this silently fails if some checks don't match up? If `user_model.USERNAME_FIELD` is set to something that isn't valid, then expected behavior wouldn't be to just silently assume a different model?

src/wiki/forms_account_handling.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+User = get_user_model()
+
+
+def check_user_field(user_model=User):
+    return isinstance(_get_field(user_model, user_model.USERNAME_FIELD), CharField)
+
+
+def check_email_field(user_model=User):
```
benjaoming: `User` is assigned as a default kwarg, but it seems to always be called w/o kwargs?

src/wiki/forms_account_handling.py
```diff
@@ -0,0 +1,89 @@
...
+    try:
+        return model._meta.get_field(field)
+    except FieldDoesNotExist:
+        return
+
+
+User = get_user_model()
+
+
+def check_user_field(user_model=User):
```
benjaoming: `User` is assigned as a default kwarg for `user_model`, but it seems to always be called w/o kwargs?

src/wiki/forms.py
trevorpeacock: [wiki/views/accounts.py](https://github.com/django-wiki/django-wiki/blob/22a432d0940189de5852ea1b07bdfe1dd83950b3/src/wiki/views/accounts.py) refers to wiki.forms.UserCreationForm

Having split some forms into forms_account_handling to make their processing more readable, I thought it would still be nice to give the appearance that forms were still in wiki.forms. The lintier complains because those imports are not used in the module.

We could alternatively update references elsewhere in the project to point to wiki.forms_account_handling instead of wiki.forms.

src/wiki/forms_account_handling.py
trevorpeacock: I'll take a closer look. I agree, I'm suspecting I didn't want to break any existing behaviour.

src/wiki/forms_account_handling.py
trevorpeacock: Its always called with the result of get_user_model() ( either with default kwarg, or in [wiki/checks.py](https://github.com/django-wiki/django-wiki/blob/68f7dc9b14478eb0370438d23de40c7e64521642/src/wiki/checks.py) also ).

Having the parameter makes it re-usable, it could check a custom user model is valid also. I'd probably prefer to make sure all calls always pass user_model, and remove the default kwargs. I'll look into doing that.

src/wiki/forms.py
trevorpeacock: what would be your preference @benjaoming ?

benjaoming: Let's declare an `__all__` in `forms.py`. This way, we won't have unintended breakage from re-organizing imports.

https://stackoverflow.com/questions/44834/can-someone-explain-all-in-python

src/wiki/forms_account_handling.py
trevorpeacock: Getting my head back around this again. Hopefully it doesn't fail silently. `tests.core.test_checks.CheckTests.test_check_for_fields_in_custom_user_model` should alert the user if there is a problem.
The reason for the `CustomUser` test is that Django/six do some pre-processing on `UserCreationForm/UserUpdateForm` at the time the class is **defined**, not used, so if there is a problem with the User model, forms_account_handling doesn't load, the above check doesn't run, and the user gets an exception from the django library, rather than a more helpful one from django-wiki checks.

Note also that `tests.core.test_checks.CheckTests.test_custom_user_model_mitigation_required` tests for the requirement for the `CustomUser` test, to alert us if django changes its behavior in the future.
This documentation accompanies the check:
```
Django & six check django.forms.ModelForm.Meta on definition, and raises an error if Meta.fields don't exist in Meta.model.
This causes problems in wiki.forms_account_handling.UserCreationForm and wiki.forms_account_handling.UserUpdateForm when a custom user model doesn't have fields django-wiki assumes.
There is some code in wiki.forms that detects this situation.
This check asserts that Django/six are still raising an exception on definition, and asserts the mitigation code in wiki.forms, and that test_check_for_fields_in_custom_user_model below are required.
```

Without this test, this exception is raised:
```
Traceback (most recent call last):
  File "./src-custom-model/manage.py", line 22, in <module>
    execute_from_command_line(sys.argv)
  File "/lib/python3.5/site-packages/django/core/management/__init__.py", line 364, in execute_from_command_line
    utility.execute()
  File "/lib/python3.5/site-packages/django/core/management/__init__.py", line 356, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/lib/python3.5/site-packages/django/core/management/base.py", line 283, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/lib/python3.5/site-packages/django/core/management/base.py", line 330, in execute
    output = self.handle(*args, **options)
  File "/lib/python3.5/site-packages/django/core/management/commands/check.py", line 68, in handle
    fail_level=getattr(checks, options['fail_level']),
  File "/lib/python3.5/site-packages/django/core/management/base.py", line 359, in check
    include_deployment_checks=include_deployment_checks,
  File "/lib/python3.5/site-packages/django/core/management/base.py", line 346, in _run_checks
    return checks.run_checks(**kwargs)
  File "/lib/python3.5/site-packages/django/core/checks/registry.py", line 81, in run_checks
    new_errors = check(app_configs=app_configs)
  File "/lib/python3.5/site-packages/django/core/checks/urls.py", line 16, in check_url_config
    return check_resolver(resolver)
  File "/lib/python3.5/site-packages/django/core/checks/urls.py", line 26, in check_resolver
    return check_method()
  File "/lib/python3.5/site-packages/django/urls/resolvers.py", line 256, in check
    for pattern in self.url_patterns:
  File "/lib/python3.5/site-packages/django/utils/functional.py", line 35, in __get__
    res = instance.__dict__[self.name] = self.func(instance)
  File "/lib/python3.5/site-packages/django/urls/resolvers.py", line 407, in url_patterns
    patterns = getattr(self.urlconf_module, "urlpatterns", self.urlconf_module)
  File "/lib/python3.5/site-packages/django/utils/functional.py", line 35, in __get__
    res = instance.__dict__[self.name] = self.func(instance)
  File "/lib/python3.5/site-packages/django/urls/resolvers.py", line 400, in urlconf_module
    return import_module(self.urlconf_name)
  File "/lib/python3.5/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 986, in _gcd_import
  File "<frozen importlib._bootstrap>", line 969, in _find_and_load
  File "<frozen importlib._bootstrap>", line 958, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 673, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 665, in exec_module
  File "<frozen importlib._bootstrap>", line 222, in _call_with_frames_removed
  File "./src-custom-model/testproject/urls.py", line 26, in <module>
    url(r'', include('wiki.urls'))
  File "/lib/python3.5/site-packages/django/conf/urls/__init__.py", line 50, in include
    urlconf_module = import_module(urlconf_module)
  File "/lib/python3.5/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 986, in _gcd_import
  File "<frozen importlib._bootstrap>", line 969, in _find_and_load
  File "<frozen importlib._bootstrap>", line 958, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 673, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 665, in exec_module
  File "<frozen importlib._bootstrap>", line 222, in _call_with_frames_removed
  File "./django-wiki/src/wiki/urls.py", line 6, in <module>
    from wiki.views import accounts, article, deleted_list
  File "./django-wiki/src/wiki/views/accounts.py", line 21, in <module>
    from wiki import forms
  File "./django-wiki/src/wiki/forms.py", line 39, in <module>
    from .forms_account_handling import UserCreationForm, UserUpdateForm
  File "./django-wiki/src/wiki/forms_account_handling.py", line 71, in <module>
    class UserUpdateForm(forms.ModelForm):
  File "/lib/python3.5/site-packages/django/forms/models.py", line 277, in __new__
    raise FieldError(message)
django.core.exceptions.FieldError: Unknown field(s) (email) specified for MyUser
```

With the test we can avoid that and produce our own
```
SystemCheckError: System check identified some issues:

ERRORS:
testapp.MyUser: (wiki.E011) testapp.models.MyUser.get_email_field_name() refers to a field that is not of type EmailField
	HINT: If you have your own login/logout views, turn off settings.WIKI_ACCOUNT_HANDLING

System check identified 1 issue (0 silenced).
```

The situation is far from ideal, containing it in forms_account_handling with that check is the best I've been able to come up with. @benjaoming please let me know what you think?

benjaoming: I had to read the explanation two times, but it was well-worth it. Thanks for explaining it. It's actually an interesting case: Having to adapt "models.py" to avoid normal exceptional states in order to raise the exceptional state from "checks.py". It seems like an unfortunate side-effect of the checks framework.

Thanks a lot, I'm good with the chosen solution. The advantage is as you say that the checks message is more helpful.

trevorpeacock:
thanks @benjaoming . You helped me learn a lot about github in the process. I saw the release on pypi, hoping to have some time experiment with it again later today.

benjaoming:
That's awesome! Contributing to open source can be a nice learning experience, and that's a great motive to get involved!

Please feel most welcome to do more PRs :+1:

PR #869, opened 2018-06-03
Update installation.rst: fix import

dingens:
`context_processors` in Django 1.10 and above has been moved from `core` to `template`.

(source: https://stackoverflow.com/a/44071228/196244 )


benjaoming:
Thanks for this correct!! :)

PR #871, opened 2018-06-09
S3 storage fix

### LINKED ISSUE #870: "This backend doesn't support absolute urls" deleting image
I will try to explain the best as I can what I think is a bug but, given my relative knowledge of django could be some misconfiguration. Also, it could be more of a conflict between django wiki and S3 as storage.

The use case is: having a production cookiecutter django, with S3 configured, and django wiki as application. In fact locally this doesn't happen so I had to bring every AWS configuration locally to make it work.

When you delete an image, you get prompted if you want to delete the image. After that the method fires to delete the image. The problem is that ImageField using S3 storage class needs to use the "name" field instead of path because path isn't implemented in the storage class.

```
def on_image_revision_delete(instance, *args, **kwargs):
    if not instance.image:
        return
    # Remove image file
    path = instance.image.path.split("/")[:-1] #path returns not implemented, while .name works.
    instance.image.delete(save=False)
```

`diango.core.files.storage.py`

```
def path(self, name):
        """
        Return a local filesystem path where the file can be retrieved using
        Python's built-in open() function. Storage systems that can't be
        accessed using open() should *not* implement this method.
        """
        raise NotImplementedError("This backend doesn't support absolute paths.")
```
I realize that my explanation is faulty and missing some basic knowledge, sorry for that, I'm currently deepening my django skills so maybe there is something basic that I might be missing. I hope this helps figuring out a way to allow using S3 as a storage.

It would be much appreciated to know if this is an issue or if I just missed something. I will post to stack overflow for help anyway but if this is an error I'd like to help fixing it.

superandrew:
This should fix #870

### DIFF
--- a/src/wiki/plugins/images/models.py
+++ b/src/wiki/plugins/images/models.py
@@ -108,7 +108,11 @@ def on_image_revision_delete(instance, *args, **kwargs):
     if not instance.image:
         return
     # Remove image file
-    path = instance.image.name.split("/")[:-1]
+    import storages
+    if isinstance(instance.image.storage, (storages.backends.s3boto3.S3Boto3Storage,)):
+        path = instance.image.name.split("/")[:-1]
+    else:
+        path = instance.image.path.split("/")[:-1]
     instance.image.delete(save=False)
 
     # Clean up empty directories
--- a/src/wiki/plugins/images/models.py
+++ b/src/wiki/plugins/images/models.py
@@ -108,7 +108,7 @@ def on_image_revision_delete(instance, *args, **kwargs):
     if not instance.image:
         return
     # Remove image file
-    path = instance.image.path.split("/")[:-1]
+    path = instance.image.name.split("/")[:-1]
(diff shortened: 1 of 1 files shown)


src/wiki/plugins/images/models.py
```diff
@@ -108,7 +108,11 @@ def on_image_revision_delete(instance, *args, **kwargs):
     if not instance.image:
         return
     # Remove image file
-    path = instance.image.path.split("/")[:-1]
+    import storages
```
benjaoming: This module doesn't exist (it's optional) -- the reason it passes tests is actually because this isn't covered by any tests :/

I think you should change this to a `try ... except` block. Catch `NotImplemented`  and then just `return` when it doesn't exist, because then there's no path to delete.

superandrew: of course, will do. Could you give me a hint on how I should write tests? Id' be happy to learn.

PR #874, opened 2018-06-10
S3 storage fix

superandrew:
Fixes S3 Storage fix for uninmplemented path method

### DIFF
--- a/src/wiki/plugins/images/models.py
+++ b/src/wiki/plugins/images/models.py
@@ -107,9 +107,14 @@ class ImageRevision(RevisionPluginRevision):
 def on_image_revision_delete(instance, *args, **kwargs):
     if not instance.image:
         return
-    # Remove image file
-    path = instance.image.path.split("/")[:-1]
+    # Remove image file    
     instance.image.delete(save=False)
+    
+    try:
+        path = instance.image.path.split("/")[:-1]
+    except NotImplemented:
+            # This backend storage doesn't implement 'path' so there is no path to delete
+        return
 
     # Clean up empty directories
 


superandrew:
@benjaoming not sure which tests have failed, seems ok to me.. 📌

benjaoming:
Hi @superandrew, thanks for responding to the review!

It's not the tests that are failing, it's doing the error because normally we shouldn't merge code that is uncovered by tests.

In this case, I'm okay with that particular exception :)

src/wiki/plugins/images/models.py
```diff
@@ -108,7 +108,12 @@ def on_image_revision_delete(instance, *args, **kwargs):
     if not instance.image:
         return
     # Remove image file
-    path = instance.image.path.split("/")[:-1]
+    try:
+        path = instance.image.path.split("/")[:-1]
+    except NotImplemented as e:
```
benjaoming: In this line, you shouldn't assign `e` since it isn't used, it's possible to merely do `except NotImplemented`.

After changing it, try using:

```
git add src/wiki/plugins/images/models.py
git commit --amend
git push -f
```

This will overwrite the previous commit and force-push to the remote branch, so you don't have to create a new PR.

Nice with the comment below :+1:

superandrew: yeah, good catch, at first I thought about logging it but there is little use of it.

benjaoming:
Thanks for fixing this! :)

src/wiki/plugins/images/models.py
```diff
@@ -108,7 +108,12 @@ def on_image_revision_delete(instance, *args, **kwargs):
     if not instance.image:
         return
     # Remove image file
-    path = instance.image.path.split("/")[:-1]
+    try:
+        path = instance.image.path.split("/")[:-1]
+    except NotImplemented:
+            # This backend storage doesn't implement 'path' so there is no path to delete
+        return
```
benjaoming: Wait, forgot one thing.. I think you need to move the whole `try ... except` block below the `delete` call. Such that the image is also deleted on S3. I would assume it supports this.

superandrew:
done! good catch @benjaoming

benjaoming:
Thanks!

PR #875, opened 2018-06-17
fix if the argument is not "depth"

teury:
if the argument is not "depth" The page is broken.

article list 1

article list 2

article succes

I just added the commit  https://github.com/django-wiki/django-wiki/pull/336/commits/fc604e68188ce550aea642e67ff54520cedf5360

### DIFF
--- a/src/wiki/plugins/macros/mdx/macro.py
+++ b/src/wiki/plugins/macros/mdx/macro.py
@@ -59,7 +59,14 @@ class MacroPreprocessor(markdown.preprocessors.Preprocessor):
                                     value = value.replace("\\", "")
                                     value = value.replace("¤KEEPME¤", "\\")
                             kwargs_dict[str(arg)] = value
-                        line = getattr(self, macro)(**kwargs_dict)
+                        try:
+                            line = getattr(self, macro)(**kwargs_dict)
+                        except TypeError:
+                            # Catch invalid args
+                            line = line[m.start():m.end()]
+                        except ValueError:
+                            # Catch invalid values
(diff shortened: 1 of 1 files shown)


Mortal:
On our wiki we use the [following helper](https://github.com/TK-IT/web/blob/ece0f5d0d43ac575af2e8a2d6f1806064893f66f/tkweb/apps/eval/evalmacros.py#L36-L42) to return a piece of error HTML:

```python
def _inline_error(method, error):
    html = (
        '<span class="tk-error">'
        + '<span class="btn btn-danger">%s</span>'
        + '<span class="btn btn-default">%s</span>'
        + "</span>"
    ) % (method, error)
    return html
```

Perhaps the macros plugin could use something similar in this case?

I'd suggest adding test cases for invoking [article_list] with wrong arguments to ensure that this PR handles all the common error cases. For test inspiration you can use the [redlinks tests](https://github.com/django-wiki/django-wiki/blob/master/tests/plugins/redlinks/test_redlinks.py) in which markdown parsing is tested.

benjaoming:
@Mortal - I think the macro plugin should error out completely if there are unhandled exceptions. Such errors are relevant for the develop/system admin, not the user :)

teury:
I think we can put an error message for the user, "Badly written text or invalid condition"

So that the user can correct the error.

benjaoming:
@teury - agreed, but there are 2 cases: Caught exceptions (bad user input), and uncaught exceptions (failing macro method). Just wanted to point that out to be sure we don't put it in the same category :)

PR #881, opened 2018-06-30
Specify extensions as module names rather than Extension instances

Mortal:
Modify `core_extensions()` and the `macros` plugin to specify extensions with their module names instead of instantiating them. This allows users to pass configuration settings to those extensions.

Also update TOC extension to simply use the `TocExtension` class in upstream Markdown. Update settings documentation to actually show how to set the title of the table of contents.

Also use new configuration API in Wikilinks extension.


src/wiki/conf/settings.py
```diff
@@ -45,9 +50,6 @@
         'markdown.extensions.abbr',
         'markdown.extensions.sane_lists',
     ],
-    'extension_configs': {
-        'toc': {
-            'title': _('Table of Contents')}},
```
benjaoming: This is a string marked for translation. Removing it also removes it from the set of translated strings -- and I dunno, is it translated in Markdown's implementation?

Or was the previous `title` config simply not rendered anywhere? (I don't see it appearing)

benjaoming:
Thanks for this! Really improves the code!

Especially getting rid of the TOC code. @atombrella had already done a bit of work for this in #739 - but this PR seems to be the full consequence. It seems that more improvements have been made to the TOC extension upstream since this code was added in django-wiki. So it's likely also an improvement.

For a bit of extra verification, we can release it as another 0.4 alpha and check that everything looks fine.
Please do also add a release note?

src/wiki/conf/settings.py
Mortal: The `title` config was not rendered since the `'toc'` key does not resolve to any extension. An alternative would be fixing the `MARKDOWN_KWARGS` so that the translated title would actually appear on django-wiki sites, but this would be a change in the behavior.

Here's a bit of code analysis showing why the `'toc'` key is not used for anything. The `extension_configs` dictionary eventually ends up in `markdown.Markdown.__init__()`:

```python
        self.registerExtensions(extensions=kwargs.get('extensions', []),
                                configs=kwargs.get('extension_configs', {}))
```

... and this leads to the following spot in `registerExtensions()` where `configs` is only used for string members of the `extensions` list:

```python
    def registerExtensions(self, extensions, configs):
        for ext in extensions:
            if isinstance(ext, util.string_type):
                ext = self.build_extension(ext, configs.get(ext, {}))
            ...  # continue using `ext` as an Extension instance
```

After following the code in this manner, I can't see how a key of `'toc'` in `extension_configs` could ever get to the `WikiTocExtension` instance created in `src/wiki/plugins/macros/wiki_plugin.py` - but maybe there's some other magic in the code I haven't found.

Mortal: Yes, that would work. I didn't do it in the PR because that's a change in behavior, but we can add titles to TOCs in django-wiki if you want.

Mortal: I've added a commit that adds the title to TOCs. While doing so I noticed that Python-Markdown doesn't like having gettext_lazy objects in the ElementTree structures, so I fixed WikiTreeProcessorClass to handle the translated title properly.

Mortal:
I've updated the release notes, extended the documentation of `MARKDOWN_KWARGS`, and added a title to the table of contents.

Mortal:
Looks good - thanks.

PR #884, opened 2018-07-10
Removes unnecessary endline/startline characters from strings

Ayub-Khan:
Removes unnecessary endline/startline characters from strings


Ayub-Khan:
@benjaoming not a big change but i hope it helps.

benjaoming:
Thanks, we are in 0.4 alpha stage - so the text changes should arrive now (because of translations), perfect timing :)

Issue #485, opened 2015-10-28
Broken interwiki links [[...]] are not highlighted


guettli:
Broken interwiki links [[...]] are not highlighted.

It would be nice if you see immediately which links are broken. 

Dokuwiki displays these links red underlined.

Is this a bug or feature?

benjaoming:
It's a feature. Please use the link plugin for this stuff as the `[[ Link ]]` notation is not a part of the official Markdown syntax nor a really operational notation for storing data about links.

Argument 1

@the-glu is working on a Move feature #461 . This enables you to move articles. We can automatically fix stuff that is made using the link plugin if we design this properly.

Argument 2

There is a whole UI for searching articles in the Sidebar while editing. This sort of stuff is possible using our own conventions.

Argument 3

This is against using `[[ ... ]]` as our own notation: It is already in use so people will expect it to work in a specific way. I would recommend to not use it at all.

benjaoming:
No rebuttals received :) Going to close this then, but thanks for coming up with the suggestion and feel welcome if you encounter other ideas or issues.

Mortal:
How about highlighting Markdown-style links `[...](...)` to other wiki pages on the same wiki? E.g. when `/foo/bar/` contains `[example](../baz)`, it would be nice that the link is colored red if `/foo/baz/` does not exist in the wiki.

PR #889, opened 2018-07-25
Add Redlinks plugin

Mortal:
Simple implementation of #485. Checks each `<a>` tag in the tree to see if the `href` points to a wiki article that does not exist; if so, the CSS class `wiki-broken` colors the link red (#ba0000, same color as Wikipedia uses).


Mortal:
Is the file `src/wiki/static/wiki/css/all.css` actually used by django-wiki anymore? It doesn't seem to be updated by `make assets`.

tests/plugins/redlinks/test_redlinks.py
```diff
@@ -0,0 +1,25 @@
...
+from wiki.core import markdown
+
+
+class RedlinksTests(RequireRootArticleMixin, TestBase):
+    def test_internal(self):
+        md = markdown.ArticleMarkdown(article=self.root_article)
+        md_text = md.convert("[Internal](.)")
+        self.assertIn("wiki-internal", md_text)
+        self.assertNotIn("wiki-external", md_text)
+        self.assertNotIn("wiki-broken", md_text)
```
benjaoming: Could you maybe add something for the parent paths here, too? Would that make sense?

benjaoming: I mean like `[Internal](../foo/bar)`

Mortal: Certainly - I've pushed a commit with more test coverage, now with tests that use a child article in addition to the root article. Please take another look.

benjaoming:
> Is the file src/wiki/static/wiki/css/all.css actually used by django-wiki anymore? It doesn't seem to be updated by make assets.

Good catch! Also, `print.css` seems to be omitted. I'm not sure what has happened here. Maybe it got lost in the `base_site.html` refactor quite a while back.

https://github.com/django-wiki/django-wiki/blob/master/src/wiki/templates/wiki/base_site.html#L12

Of course, these declarations are easy to add in CSS sources using the `@media` rule.

I think that both `all.css` and `print.css` should be deleted. Looking in `src/wiki/static/wiki/bootstrap/less/wiki/wiki.less`, it seems that the code was moved there and has been updated since.

benjaoming:
Some really nice work there @Mortal - thanks for adding this great new and very useful feature!

PR #894, opened 2018-07-30
Add tests for buggy translation behavior

Mortal:
django-wiki contains two instances of

```django
{% blocktrans %}...{% plural %}...{% endblocktrans %}
```

... and the translations of these blocks are not in use for some reason.

On my Django wiki instance with the global history plugin enabled and da-DK locale, the strings "List of 685 changes in the wiki." and "Be careful, links to this article and 0 articles nested in this hierarchy will not be updated." are not translated.

In Transifex the strings are correctly translated, e.g. the first is translated into "Liste af %s ændringer i wikien.", and this translation is present in the installed translation file as verified with grep:

```
$ grep 'i wikien' .venv/lib/python3.5/site-packages/wiki/locale/da/LC_MESSAGES/django.mo 
Binary file .venv/lib/python3.5/site-packages/wiki/locale/da/LC_MESSAGES/django.mo matches
```

I don't know why the translations are not found by Django when rendering the templates. This PR adds two unit tests verifying the buggy behavior.


benjaoming:
Thanks for flagging this :)

> I don't know why the translations are not found by Django when rendering the templates. This PR adds two unit tests verifying the buggy behavior.

Would suggest this:

 * Commit 1: Add a test case that BREAKS, show-casing the behavior and ensuring that the test catches what you want
 * Commit 2: Adds changes to the application, fixing the test case

Mortal:
> Commit 2: Adds changes to the application, fixing the test case

How would you suggest we fix the issue? I added the test case in this way because I don't know how to fix the issue (and I don't want to commit a failing test case).

benjaoming:
Ah, haha, no, I thought you already knew :smiley:

Mortal:
I've investigated some more. I think the issue is that `makemessages` has not been run in the repo since `trimmed` was added in #884 ("Removes unnecessary endline/startline characters from strings", 2018-07-11). The `locale/en/LC_MESSAGES/django.po` file currently in the repo matches what I find when I download the source file from Transifex (<https://www.transifex.com/django-wiki/django-wiki/core/en/download/for_use/>):

```
#: templates/wiki/move.html:23
#, python-format
msgid ""
"\n"
"          %(cnt)s nested article will also be moved.<br>\n"
"          Be careful, links to this article and %(cnt)s article nested\n"
"          in this hierarchy will not be updated.\n"
"          "
msgid_plural ""
"\n"
"          %(cnt)s nested articles will also be moved.<br>\n"
"          Be careful, links to this article and %(cnt)s articles nested\n"
"          in this hierarchy will not be updated.\n"
"        "
```

However, when I run makemessages locally and inspect the file, the spaces are gone:

```
#: plugins/globalhistory/templates/wiki/plugins/globalhistory/globalhistory.html:14
#, python-format
msgid "List of <strong>%(cnt)s change</strong> in the wiki."
msgid_plural "List of <strong>%(cnt)s changes</strong> in the wiki."
```

@benjaoming If you agree with my analysis, would you kindly run `makemessages` and upload to Transifex?

benjaoming:
@Mortal I agree with that, it's a blunder from my side. When I was pushing messages last time, I didn't see that they needed to be generated - I thought they just hadn't been pushed :) Thanks and sorry! They're in master now!

Mortal:
Thanks @benjaoming - I've updated the Danish translation. Would you mind pulling from Transifex?

benjaoming:
Alright, Danish translation is in `master` :+1:

benjaoming:
Hi @Mortal - can you merge in master and negate the test to confirm that it works? I think the test case is okay to keep, even though we solved the issue :)

benjaoming:
Thanks @Mortal !

PR #896, opened 2018-07-30
Allow escaping [image] and [article_list]

Mortal:
This allows escaping the macros parsed by the macro library.

### DIFF
--- a/src/wiki/plugins/images/markdown_extensions.py
+++ b/src/wiki/plugins/images/markdown_extensions.py
@@ -3,19 +3,18 @@ from django.template.loader import render_to_string
 from wiki.plugins.images import models, settings
 
 IMAGE_RE = (
-    # Match only at the beginning of a line
-    r"(?:(?im)^\s*" +
+    r"(?:(?im)" +
     # Match '[image:N'
     r"\[image\:(?P<id>[0-9]+)" +
     # Match optional 'align'
     r"(?:\s+align\:(?P<align>right|left))?" +
     # Match optional 'size'
     r"(?:\s+size\:(?P<size>default|small|medium|large|orig))?" +
-    # Match ']' at end of line
-    r"\s*\]\s*$" +
-    # Match zero or more caption lines, each indented by four spaces.
+    # Match ']' and rest of line.
     # Normally [^\n] could be replaced with a dot '.', since '.'
     # does not match newlines, but inline processors run with re.DOTALL.
+    r"\s*\](?P<trailer>[^\n]*)$" +
+    # Match zero or more caption lines, each indented by four spaces.
     r"(?P<caption>(?:\n    [^\n]*)*))"
 )
 
@@ -62,6 +61,7 @@ class ImagePattern(markdown.inlinepatterns.Pattern):
             pass
 
         caption = m.group("caption")
+        trailer = m.group('trailer')
 
         caption_placeholder = "{{{IMAGECAPTION}}}"
         width = size.split("x")[0] if size else None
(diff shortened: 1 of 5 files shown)


benjaoming:
Thanks! That should fix an important drawback of this plugin, because it does really easily leak if someone implements the wrong patterns.

I'm a bit +0 about the `macros` plugin: It seems to help people get small features implemented in a simpler way than to write a full plugin.

benjaoming:
Maybe this can more easily fix what #875 tries to, @Mortal ?

Mortal:
I added the test to the wrong file. I'll work some more on this PR.

Mortal:
@benjaoming I completed the tests and also refactored the ImagePreprocessor as well. New commit message:

```
Allow escaping [image] and [article_list]

Change ImagePreprocessor and MacroPreprocessor into inline processors.
Preprocessors run before processing of backticks, which means that the
two tags would be handled by the images and macros plugins before the
backtick processor would escape the contents.

Add tests to ensure that escaping [image] and [article_list] works as
expected.
```

src/wiki/plugins/images/markdown_extensions.py
```diff
@@ -1,106 +1,84 @@
...
+                "caption": caption_placeholder,
+                "align": alignment,
+                "size": size,
+                "width": width,
+            },
+        )
+        html_before, html_after = html.split(caption_placeholder)
+        placeholder_before = self.markdown.htmlStash.store(html_before, safe=True)
+        placeholder_after = self.markdown.htmlStash.store(html_after, safe=True)
+        return placeholder_before + caption + placeholder_after
```
benjaoming: This is such a great example of how doing things "the right way" is rewarding: Readability, simplicity, performance etc..

tests/plugins/images/test_markdown.py
```diff
@@ -8,9 +8,17 @@ def test_markdown(self):
         md_text = md.convert("[image:1 align:left]")
         self.assertIn("<figure", md_text)
         self.assertNotIn("[image:1 align:left]", md_text)
+        md_text = md.convert("image: [image:1 align:left]\nadasd")
+        self.assertIn("<figure", md_text)
+        md_text = md.convert("image: [image:1 align:right medium]\nadasd")
+        self.assertIn("<img src", md_text)
```
Mortal: I think when I tried checking for `<img` in `md_text`, it failed because the output contained "Image not found" instead. Is there another mixin class for adding a test image?

Mortal:
Ah, I misunderstood the previous behavior of `[image]`. I see that it was not required that the `[image]` is at the beginning of the line. I'll fix that.

Mortal:
@benjaoming I've pushed some more commits to handle text before and after `[image]` on the same line properly. Please take a look.

benjaoming:
Hi @Mortal - great work! Just wanna check if captions work :)

benjaoming:
@Mortal I'm wondering what the best outcome here is?

```
[image:123]
    This is a **caption**
```

Should it be with or without the paragraph?

```
<figure>
<a href="..."><img src=""></a>
<figcaption><p>This is a <b>caption</b></p></figcaption>
</figure>
```

I'm also thinking that maybe the "image not found" part is a bit bork'ed... like it seems really cool to show the caption, even though the image is somehow gone (makes it easier to understand what's supposed to be there).

```
<figure class="thumbnail pull-left"><a href=""><div class="caption"><em>Image not found</em></div></a><figcaption class="caption">Caption text still visible</figcaption></figure>
```

1. `<a href="">` - linking to nothing
1. `<div class="caption">` around what is essentially not the caption, followed by a real caption below.

```
<figure class="thumbnail pull-left"><div class="alert alert-danger image-404" role="alert"><em>Image not found or error generating thumbnail</em></div><figcaption class="caption">Caption text still visible</figcaption></figure>
```

With something like `.image-404 {max-width: 250px;}`

Mortal:
I've fixed the tests now: In setUp(), a test image is created. I've also included a test that says that "Image not found" should still include the caption.

> Should it be with or without the paragraph?

If you mean `<figcaption><p>...</p></figcaption>` vs. `<figcaption>...</figcaption>`: Both are allowed (figcaption [permits](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/figcaption) any [flow content, including `<p>`](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Content_categories#Flow_content)), so if the implementation is simplest by including the `<p>`, then I think that's fine.

> I'm also thinking that maybe the "image not found" part is a bit bork'ed...

I think we should address that in a follow-up PR.

PR #900, opened 2018-08-04
[RFC] Combine "Search from current article" and "Search whole wiki"

Mortal:
Currently, the search field at the top of every wiki article only searches articles below the currently viewed article. On my wiki it is rarely useful to search under the current; it would be preferable that the search page shows results from the entire wiki.

I'd suggest merging "Search from current article" and "Search whole wiki" so that the search results page displays first the results under the current article, and then displays results from the whole wiki.

I'll work on an implementation in this PR.

### DIFF
--- a/src/wiki/views/article.py
+++ b/src/wiki/views/article.py
@@ -4,6 +4,7 @@ from urllib.parse import urljoin
 
 from django.contrib import messages
 from django.contrib.auth.decorators import login_required
+from django.core.paginator import InvalidPage
 from django.db import transaction
 from django.db.models import Q
 from django.http import Http404
@@ -11,11 +12,11 @@ from django.shortcuts import get_object_or_404, redirect, render
 from django.urls import reverse
 from django.utils.decorators import method_decorator
 from django.utils.translation import gettext as _, ngettext
+from django.views.decorators.clickjacking import xframe_options_sameorigin
 from django.views.generic import DetailView
 from django.views.generic.base import RedirectView, TemplateView, View
 from django.views.generic.edit import FormView
 from django.views.generic.list import ListView
-from django.views.decorators.clickjacking import xframe_options_sameorigin
 from wiki import editors, forms, models
 from wiki.conf import settings
 from wiki.core import permissions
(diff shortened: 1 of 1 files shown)


benjaoming:
That sounds interesting!

Btw you can merge querysets like this:

```
qs = models.SomeModel.objects.filter(...) | models.SomeModel.objects.filter(...)
```

This is the same as `SELECT * FROM ... UNION SELECT * FROM ...`.

benjaoming:
And I think it would work great by for instance combining with `unique()` such that there would be two headers:

> ## Results in '`{{ article.title }}`'
> 
> ...
> 
> ## Results from remaining wiki
> 
> ...

benjaoming:
Hey @Mortal og @neic - any updates on this? Would love to get it merged in, just worried about the situation with duplicate stuff. For instance if you search the "root" article and then you get confusing results from "root" and the whole wiki at once. So would be nice to combine two different querysets and somehow be able to remove duplicates from the "whole wiki" set such that it's actually "other results from the rest of the wiki".

benjaoming:
I think it is a nice change, please feel encouraged to have time :)

Maybe we could make use of the already existing structure and just offer the user a link to jump back and forth between to different search types?

```
/_search/?q=test  # Searches full wiki
/slug-name/_search/?q=test  # Searches inside 'slug-name'
```

So, currently by default if you are viewing `slug-name`, then you will search in that hierarchy. This isn't desirable!

If we add the originating article as a querystring parameter, we could retain where the user was coming from?

```
/_search/?q=test&from=/slug-name  # Searches full wiki, providing a link to only search in 'slug-name'
```

In the case of displaying a link, we could perform a count on the search results. E.g. "View search results ONLY in Article Title (123 results)".

This is easy to implement, because it doesn't require considering the size of the result query sets or to have to combine them or merge them in the pagination :)

benjaoming:
Hi @Mortal - thanks for starting the work on this. I suppose you didn't have time to finish it. You're always welcome to re-open the PR or start a new one in case you want to continue this effort!

PR #904, opened 2018-08-05
Show first match instead of last in search results

Mortal:


benjaoming:
Originally, I recall that this tried to put the matched keyword in the middle of a search result.

For instance, searching for **needle** in

> _hay hay hay needle hay hay hay_

would produce with a 3 word snippet:

> _hay **needle** hay_

But I remember also that there were some bugs, and eventually it got moved around.

PR #905, opened 2018-08-05
Updates pull in from Transifex

benjaoming:
@Mortal contains latest Danish changes.

Unfortunately after pushing messages, Transifex updates the "last activity" stamp of all languages, so it was impossible to tell which language had actual changes.

I only know for a fact that Danish, Chinese (ZH) and Japanese were updated.


Mortal:
With Git it is possible to diff the binary message catalog files as follows:

```
git config diff.msgunfmt.textconv msgunfmt
echo '*.mo diff=msgunfmt' >> .git/info/attributes
git log -p src/wiki/locale/*/LC_MESSAGES/django.mo
```

In this way I have verified that indeed only da, zh_CN and ja were updated.

benjaoming:
Thanks @Mortal - that's some magic stuff :)

I'm still pondering if maintaining the .po for all of the languages is perhaps better, since it'll make it clear for each change. Perhaps, it's even better for git storage, as the delta's might be smaller for text files than binaries.

PR #909, opened 2018-09-15
Allow Django 2.1 and add to tests

olymk2:


setup.py
```diff
@@ -24,7 +24,7 @@ def get_path(fname):
 
 
 install_requirements = [
-    "Django>=1.11,<2.1",
+    "Django>=1.11,<2.1.1",
```
benjaoming: This comes with the same issues that I mentioned in https://github.com/django-wiki/django-nyt/pull/82

We cannot allow Django 2.1 without adding it to the test matrix, fixing whatever issues come along.

benjaoming:
Could you give an example of the warnings you mentioned in the subject?

PR #910, opened 2018-09-18
Fix typo in name

Mortal:


benjaoming:
Garh, that's why I shouldn't type names and constants manually. Always copy-paste :)

PR #912, opened 2018-10-15
Fix image deletion issue caused by S3 storage fix

### LINKED ISSUE #911: [Bug] NotImplemented is not an exception
https://github.com/django-wiki/django-wiki/blob/a144a3110cc688d796895aa1caddc4db1a4b62c6/src/wiki/plugins/images/models.py#L122

I think this is a bug as 'NotImplemented" is not an exception but a singleton. Delete lines 120-124?

[Docs](https://docs.python.org/3.6/library/constants.html#NotImplemented)

benjaoming:
Fixes #911 

This fixes some code that may have been introduced through a merging issue. I don't know -- but it addresses something that was assumed fixed and is already in the changelog, so will not add it again :)

### DIFF
--- a/src/wiki/plugins/images/models.py
+++ b/src/wiki/plugins/images/models.py
@@ -110,16 +110,10 @@ def on_image_revision_delete(instance, *args, **kwargs):
 
     # Remove image file
     instance.image.delete(save=False)
-    
-    try:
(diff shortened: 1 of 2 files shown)


benjaoming:
Bug reported to CodeCov yesterday, but I give up about the comment not being updated.

PR #915, opened 2018-10-16
If read isn't allowed, then display back button to parent article

### LINKED ISSUE #914: "Back to article" button links to itself on protected articles
It works like expected on special pages like `_create`, but not on wiki articles the user doesn't have access to. 
For example https://demo.django.wiki/HiddenArticle/ links to `/HiddenArticle/`. 
It should probably link to `../` or not display the back button if that doesn't make sense.

benjaoming:
Fixes #914

### DIFF
--- a/src/wiki/templates/wiki/permission_denied.html
+++ b/src/wiki/templates/wiki/permission_denied.html
@@ -17,20 +17,16 @@
   </p>
   {% endif %}
 
-  {% if not read_denied %}
   <p>
+  {% if not read_denied %}
     <a href="{% url 'wiki:get' article_id=article.id path=urlpath.path %}" class="btn btn-default">
       {% trans "Back to article" %}
     </a>
-  </p>
-  {% else %}
-  <p>
-    {% if urlpath.parent %}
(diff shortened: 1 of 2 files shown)


src/wiki/templates/wiki/permission_denied.html
```diff
@@ -17,10 +17,20 @@ <h1 class="page-header">{% trans "Permission Denied" %}</h1>
   </p>
   {% endif %}
 
+  {% if not read_denied %}
   <p>
     <a href="{% url 'wiki:get' article_id=article.id path=urlpath.path %}" class="btn btn-default">
       {% trans "Back to article" %}
     </a>
   </p>
+  {% else %}
```
duvholt: Instead of nesting else and if I think an elif would be better. 

```django
  {% elif urlpath.parent %}
    <p>
      <a href="{% url 'wiki:get' path=urlpath.parent.path %}" class="btn btn-default">
        {% trans "Back to article" %}
      </a>
    </p>
  {% endif %}
```

PR #916, opened 2018-10-17
Docs updates for Django 2

benjaoming:
For the next release, these installation instructions would cause less confusion.


benjaoming:
@duvholt would you have time to give a quick look at this? :)

benjaoming:
After this, I think we could just release 0.4 stable. Not sure if there are other bigger issues to solve?

Current milestone issues are features/improvements that don't block Django 2.x usage... https://github.com/django-wiki/django-wiki/issues?q=is%3Aopen+is%3Aissue+milestone%3A0.4

docs/installation.rst
```diff
@@ -155,14 +155,18 @@ point to the root article:
 Include urlpatterns
 ~~~~~~~~~~~~~~~~~~~
 
-To integrate the wiki to your existing application, you should add the
-following lines at the end of your project's ``urls.py``.
+To integrate the wiki to your existing application, you should ensure the
+following lines in your project's ``urls.py``.
```
duvholt: Nitpick, but I think this reads a bit odd.

How about:
>To integrate the wiki in your existing application, you should ensure the following lines are included in your project's ``urls.py``. 

My English isn't perfect so this might not be any better. :upside_down_face:

benjaoming: Thanks, you are absolutely right :)

Issue #882, opened 2018-07-05
jquery 1.12.4 is bundled, some security holes exist


donbowman:
see [CVE-2015-9251](https://cve.mitre.org/cgi-bin/cvekey.cgi?keyword=CVE-2015-9251) and others

benjaoming:
Thanks! Even worse is that for some perceived convenience I removed the version string from the file name. Which makes the problem harder to spot.

https://github.com/django-wiki/django-wiki/tree/master/src/wiki/static/wiki/js

Quick work-around: Put a fixed version of JQuery in a static path `wiki/js/jquery.min.js`, for instance by using `STATIC_DIRS`.

benjaoming:
Fixed in #919 - sorry about the delays here! This issue should have had a milestone assigned.

PR #918, opened 2018-10-20
Update jquery to 3.3.1

donbowman:
This resolves https://github.com/django-wiki/django-wiki/issues/882, specifically that jquery is old and has vulnerabilities including CVE-2015-9251.

Signed-off-by: Don Bowman <db@donbowman.ca>


benjaoming:
Sorry, I can't accept a minified jquery due to the same reasons of security :)

But could you tell me if you have tested whether this works?

benjaoming:
And thanks for flagging this again!

benjaoming:
Ah yes, thanks - no what I meant was also that making the comparison is essentially the same as adding it myself :)

I've opened another PR and done some tests here and there of Editing, preview functions etc.. seems like the editor and modal dialogs are intact.

PR #920, opened 2018-10-20
Add support for markdown >= 3.0

donbowman:
[python-markdown](https://github.com/Python-Markdown/markdown) is now
at version 3.0.1. Some interfaces that were formerly deprecated
are now removed. Make corresponding changes here, notably:

- remove 'smart_strong'
- remove 'safe_mode' HTML (was not implemented)
- remove 'md_globals'
- fix positional argumenst (e.g. extensions=[])
- support InlineProcessor (instead of Pattern) in WikiLinks

### DIFF
--- a/requirements_readthedocs.txt
+++ b/requirements_readthedocs.txt
@@ -5,5 +5,5 @@ Django>=1.8,<1.12
 Pillow
 django-sekizai>=0.10
 sorl-thumbnail>=12,<13
-Markdown>=2.6,<2.7
+Markdown>=3.0.0,<3.1.0
 bleach>=2.1,<2.2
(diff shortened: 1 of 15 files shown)


benjaoming:
Thanks! This is most welcome!

With this level of coverage, this change looks totally safe!

It would, however be necessary to bump the minor version to 0.5 for a major dependency version bump like this. But I don't have any bureaucratic issues with doing that :)

donbowman:
it affects the interop with the pymdownx and the uml sequence markdown that
i use.

donbowman:
i'm not clear, are you asking me to bump the ver to 0.5.0 on this branch?

PR #924, opened 2018-10-25
Simplification of message tag resolution and setting, fixes #922

### LINKED ISSUE #922: sekizai error python 3.6
It is to report what looks like an error in python 3.6
sekizai, has stopped updating, which is a bad idea to use this long-term dependency.

benjaoming:
@dexbleeker @teury - would you have time to review/test this? Thanks :)


benjaoming:
With a diff coverage of 100% and a msg from @teury that it works, I'll merge+release this.

Issue #927, opened 2018-11-06
Urls are matched as 0.123


benjaoming:
The FQDN TLD group is allowed to be digits only, which makes for instance `0.123` turn into a link.

![urlize](https://user-images.githubusercontent.com/5114048/48092961-723a0c00-e1e4-11e8-8113-d1a55122b1d4.jpg)

I see a small error that creates links after a point.

_Originally posted by @teury in https://github.com/django-wiki/django-wiki/pull/816#issuecomment-436435331_

benjaoming:
Nice @teury - all 3 would be important fixes. Thanks!

benjaoming:
CTRL+Enter just closed the issue... hmm.. _feeling confused_

benjaoming:
Fixed in #937

PR #928, opened 2018-11-09
Fix the numbers are detected as TLD domains

teury:
1. The error causes the numbers to be detected as a domain (TLD)
2.  I have seen that links are created by typing the word "localhost" This is a very frequent word, and in most occasions a redirect is not necessary.

#927


benjaoming:
Hi @teury - would you be able to get back with a fix of the tests that

1) Ensures that for instance `http://localhost` is still highlighted?
1) Deletes the current `localhost` test, because this is of course the one we don't want highlighted.

teury:
the test fails because it verifies that the link is created in localhost.
we just have to modify the test

Issue #930, opened 2018-11-15
Article cache breaks with Memcached when User.__str__ contains whitespace.


oleast:
## Problem 
Per user cache for articles uses the `__str__` method of the User to create a key for the user. [link](https://github.com/django-wiki/django-wiki/blob/master/src/wiki/models/article.py#L208-L212)

The Django [docs](https://docs.djangoproject.com/en/2.1/ref/models/instances/#str) encourages the use of a human readable string `__str__`. Our project uses the given example of:
```
'%s %s' % (self.first_name, self.last_name)
```
As it is quite helpful in our use-case to have the full name of our users as the default. I can image a lot of other projects also using the full names of their users, with the whitespace.

Memcached does not allow whitespace in its keys, which will result in the entire wiki returning a server error to a user if their `__str__` contains whitespace.

This means that any project using this configuration will break.
I don't know if there are many others with the same configuration, but I can imagine that we are not alone.

## Possible solution

To solve this problem we could utilize the the `get_username()` method of User, instead of the `__str__` method. This is the same method which is used by AbstractBaseUser for its `__str__` [source](https://github.com/django/django/blob/master/django/contrib/auth/base_user.py#L66-L67).

This will also have the added benefit of making sure the cache key for each individual user is unique, which `__str__` may not be in all cases, depending on the project.

benjaoming:
I agree with your analysis and solution. Thanks for an elaborate explanation. Too bad about memcached, but of course one cannot just automatically replace whitespaces in the keys such that `user name` becomes the same as `user_name`.

Whitespaces aren't allowed in Django's username field hopefully then :)

oleast:
Memcached only disallows whitespace characters and control characters. [Docs](https://github.com/memcached/memcached/blob/master/doc/protocol.txt#L41-L49)

Django uses a quite simple regex for allowed characters:
```
regex = r'^[\w.@+-]+$'
```
[Source](https://github.com/django/django/blob/master/django/contrib/auth/validators.py#L19-L25)

If I'm not terribly mistaken there should not be a problem with how things currently work in django 😄

benjaoming:
Thanks for checking that @oleast !

benjaoming:
From what we can gather here, all whitespace characters should be disallowed in the per user cache key, and memcached will work.

This should be easy to write a fix for and a regression test..

oleast:
It seems that I may have forgotten to directly link this issue to the PR I made (#931) 

[...]

PR #931, opened 2018-11-15
Use User.get_username() for article cache instead of User.__str__

oleast:
Detailed description in #930 

Base the caching key on the username of a User instead of the `__str__`, to increase compatibility with Django projects using Memcached as their caching service.

The default behavior of `__str__` is to use `get_username()`, but many projects choose to overwrite the method for their custom users to show the users full name.


benjaoming:
Cherry-picked and released as 0.4.4

PR #932, opened 2018-12-04
Remove localhost in the tests

teury:
I have reversed the logic so that the positive becomes negative


benjaoming:
Could you add this to your other PR?

I see that the branch is called `patch-4` -- are you writing code using Github's interface?

PR #953, opened 2019-06-09
Removed Python 3.4 from test matrix.

atombrella:


benjaoming:
Thanks, makes total sense. Heads up @jchristgit on #951

PR #970, opened 2019-09-25
[WIP] Update version table and testing matrices

atombrella:
`travis.yml` has Django 2.2 in it, so I guess it's left out by mistake?

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,6 +1,6 @@
 [tox]
 # Ensure you add to .travis.yml if you add here, using `tox -l`
-envlist = {py35,py36}-django{111,20,21,22},py37-django{20,21,22},lint
+envlist = {py35,py36,py37}-django{21,22},{py36,py37,py38}-django{30},lint
 
 [travis]
 python =
@@ -10,10 +10,8 @@ python =
 
 [travis:env]
 DJANGO =
-  1.11: django111
-  2.0: django20
-  2.1: django21
   2.2: django22
+  3.0: django30
 LINT =
   yes: lint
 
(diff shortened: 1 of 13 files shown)


benjaoming:
Thanks @atombrella 

Release 0.4.x doesn't support 2.2 officially. Current master targets 0.5.

See: https://github.com/django-wiki/django-wiki/blob/releases/0.4.x/.travis.yml

atombrella:
@benjaoming OK. Can I add the `0.5 (master)` line to the table?

tox.ini
```diff
@@ -34,16 +31,13 @@ usedevelop = false
 
 deps =
   .[test]
-  django111: Django>=1.11,<2.0
-  django20: Django>=2.0,<2.1
-  django21: Django>=2.1,<2.2
   django22: Django>=2.2,<2.3
+  django20: Django>=3.0a1,<3.1
```
benjaoming: Should say `django30`

benjaoming:
Maybe supporting 2.1 is okay? It's supported until end of 2019 and may provide a nice overlap for upgrading (so you don't have to upgrade Django AND django-wiki at the same time).

tox.ini
atombrella: Yes, I noticed myself :) I'll push some updates today, and we'll see how it works. It seems that no additional work needs to be done to support Django 3.0. I might be wrong, though.

atombrella:
> Maybe supporting 2.1 is okay? It's supported until end of 2019 and may provide a nice overlap for upgrading (so you don't have to upgrade Django AND django-wiki at the same time).

Perhaps. What's the timeline for Django Wiki 0.5? What do you think about Python 3 only for Django Wiki 0.5?

atombrella:
1.11 is LTS, 2.2 is LTS, so there's probably a good chance that only having official support for the two newest Django releases works.

benjaoming:
> Perhaps. What's the timeline for Django Wiki 0.5?

I haven't recorded anything grave lately, so I guess shipping it after this change is a good possibility!

You have any opinions or inputs on that decision?

atombrella:
No. Shipping soon would be good. Then Django 2.1 makes sense.

PR #978, opened 2019-10-07
Fixed some semantic/syntax errors in the notifications plugin.

atombrella:


src/wiki/plugins/notifications/views.py
```diff
@@ -40,14 +40,12 @@ def get_article_subscriptions(self, nyt_settings):
...
-        kwargs = super().get_form_kwargs(self)
+        kwargs = super().get_form_kwargs()
         kwargs['user'] = self.request.user
         return kwargs
 
     def get_context_data(self, **kwargs):
-        context = super().get_context_data(self, **kwargs)
-        if 'form' not in context:
-            context['form'] = self.get_form()
+        context = super().get_context_data(**kwargs)
```
benjaoming: Did this fix a syntax error?

atombrella: No, I'll take it. It's dead code, but since I have a WIP PR #979 already open to add some tests for the notifications plugin, that change could go there instead.

benjaoming: :+1:

benjaoming:
Yeah, sounds cool to add the fix to the other PR if it's going to  be matched up with a test there :+1:

benjaoming:
Fixes #1000

PR #979, opened 2019-10-07
Added some tests for notifications plugin.

atombrella:


setup.py
```diff
@@ -84,8 +84,10 @@ def get_path(fname):
         'Intended Audience :: Developers',
         'Operating System :: OS Independent',
         'Programming Language :: Python',
+        'Programming Language :: Python :: 3 :: Only',
         'Programming Language :: Python :: 3.5',
         'Programming Language :: Python :: 3.6',
+        'Programming Language :: Python :: 3.7',
```
atombrella: Maybe we should also add Python 3.8 to the list, and to the test matrix in Travis?

benjaoming: Yes, that would be great, I can see you're working on it in a separate PR now :tada:

benjaoming:
This could need a basic check that notifications are created when some article is changed.

PR #983, opened 2019-10-22
Removed unneeded/dead import fallback in wiki.compat.

atombrella:


atombrella:
@benjaoming I second that opinion and just did a force-push.

PR #984, opened 2019-10-23
Removed wiki.compat. Functionality is in django.templates.Engine.

atombrella:


atombrella:
@benjaoming Yes, indeed. I just updated.

Issue #985, opened 2019-10-23
re_path alised to url


benjaoming:
Since the cleanup in https://github.com/django-wiki/django-wiki/pull/983, we accepted to use an alias that's not desirable for the longer term.. it's a bit confusing to people that don't know the history of Django before version 2.

The issue is to use `from django.urls import re_path` and correct the code. In some places, we might be able to use just `path`.

tomasmor42:
Can I pick this up?

benjaoming:
Fixed in #986 - thanks for the swift initiative @tomasmor42

PR #986, opened 2019-10-27
Re path alised

tomasmor42:
#985


benjaoming:
Fantastic work @tomasmor42 - thanks for helping to keep the codebase healthy :)

PR #990, opened 2019-11-08
Bump to Markdown 3.1 series

### LINKED ISSUE #989: Markdown version bump needed for rest framework
restframework requires markdown 3.1.1 resulting in this error with restframework.

https://stackoverflow.com/questions/57479412/attributeerror-at-ordereddict-object-has-no-attribute-register-in-django-r

benjaoming:
Fixes #989

### DIFF
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -17,7 +17,7 @@ Release plan
 Changed
 ~~~~~~~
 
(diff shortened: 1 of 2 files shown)


benjaoming:
Since the tests are passing, I will go ahead and assume that this is fine...

PR #1003, opened 2020-01-02
Remove unused constant

citxx:
The usage was removed in https://github.com/django-wiki/django-wiki/commit/c9b7f0b7b370c21b52f8e6c937687e9edb1d748b#diff-916cdb5c7b700ddd753057b9b0945e86


benjaoming:
Thanks, that makes total sense!

PR #649, opened 2017-06-19
[WIP] Global history tweaks

benjaoming:
Fixes #648

PR #1009, opened 2020-01-07
Fixed some lint warnings, ignored E128 (temporarily)

atombrella:


benjaoming:
This looks good, but I fear there are many more places with somewhat odd/hard to read automatic code formating (from a previous tool) - I was actually thinking that having cleaned up all PRs (except #649) then it might be time to introduce Black.

The concept could be to Black format the whole codebase with 0 manual touches, such that it's 100% reproducible = verifiable despite the huge changeset.

benjaoming:
@atombrella black is now part of the pre-commit hooks and I'll add it to Travis.

It seems also that the LINT=yes environment got missing from .travis.yml

Opened: https://github.com/django-wiki/django-wiki/pull/1016

PR #1011, opened 2020-01-09
Add black and lint full repo (reproducible!)

benjaoming:
This reformats the whole repo but in a reproducible way.

It can be verified like this:

1) Checkout a master branch
2) Cherry-pick eff327f
3) Run `precommit install` and `pre-commit run --all-files`
4) Verify that you have the same result as  254c4a2 

I see a lot of valuable code reformatting going on, especially cleaning up odd linebreaks in long string expressions.


benjaoming:
Since tests pass and coverage is unchanged, I think we can merge this.

Waiting for you @atombrella

CC: @davidzwa

benjaoming:
This is a sensitive PR that can't linger, because of all the potential merge conflicts that it can create.

If anyone is concerned about the security aspect of such a large changeset: It's reproducible (with black 19.10).

PR #1012, opened 2020-01-12
Adds docker container for test_project

davidzwa:


davidzwa:
Might improve or move into separate folder!

davidzwa:
Very well, will look into it at a later moment.

benjaoming:
Oh, there is also a potential to add a sentence about it here: https://github.com/django-wiki/django-wiki#docker-tldr

benjaoming:
Hi @davidzwa - I'm not sure if the Dockerfile will contribute much as is. It should possibly be running a later version of Python, and it would need maintenance on its own.

I can understand why people would want to Dockerize the development, though -- yet at the same time, it's possible that they're better off writing their own Dockerfile tailored for their needs?

LMK if this should be reopened, this is not a "no" to Docker.

davidzwa:
It's fine! I shouldve spent more time adjusting the file. I do think dockerizing is nice, but then 100% and not like this partially as I did xD

PR #1014, opened 2020-01-16
Fix can't delete Article with USE_THOUSAND_SEPARATOR = True #1013

### LINKED ISSUE #1013: Can't delete Article with USE_THOUSAND_SEPARATOR = True
When USE_THOUSAND_SEPARATOR in django settings is set to True, deleting an article yields in an error bacause the form partly fails validation and the revision id is missing in cleaned data. This is because the id set in the hidden field in wiki/templates/wiki/delete.html is automatically formatted (e.g. 1.333 instead of 1333).
When modifying the template like this:
`{{ article.current_revision.id|safe }}`
deletion is possible.

tim3towers:
Fixes #1013


benjaoming:
Thanks so much for this @tim3towers 

Will ping back once released.

PR #1016, opened 2020-01-23
Linting in Travis & lint

benjaoming:
Related:

#1011 
#1009


benjaoming:
This PR isn't yet functional, somehow `LINT=yes` set in `.travis.yml` isn't properly handled by `tox.ini`

tox.ini
```diff
@@ -5,12 +5,13 @@ envlist = {py35,py36,py37}-django{21,22},{py36,py37,py38}-django{30},lint
...
 python =
   3.5: py35
-  3.6: py36
+  3.6: py36, lint
   3.7: py37
   3.8: py38
 
 [travis:env]
 DJANGO =
+  2.1: django21
```
atombrella, line 14: Django 2.1 is EOL. So drop this line?

benjaoming: It's not actually that line that triggers the build, that's sorted in `.travis.yml`

I had to add it to make it explicit that the py36-django21 environment should not include `lint`. It's a bit hard to explain, but it's here: https://tox-travis.readthedocs.io/en/stable/envlist.html#advanced-configuration

PR #1020, opened 2020-02-06
Update Bleach version

slinkymanbyday:
Update bleach to version 3


benjaoming:
This seems like a really good idea, thanks!

Issue #1024, opened 2020-02-15
Django 3.0 for v0.6 (non-dev)


davidzwa:
Now that Django 3.0 is out, we should start looking into getting into release of django-wiki (also ASGI and asyncio: very nice stuff)

[Edit]: noticed it is already being working on in 0.5/0.6

benjaoming:
HI @davidzwa - does 0.6b1 look good?

PR #1025, opened 2020-02-15
Bump to 0.6b1

benjaoming:
Trying a beta release firstly...


benjaoming:
Ref https://github.com/django-wiki/django-wiki/issues/1024

Issue #936, opened 2018-12-20
Deleting images fails


humberto77:
Deleting image crashes for call to `path = instance.image.path` due to incorrect exception handling.  Editing line 116 of `/src/wiki/plugins/images/models.py` to handle a ValueError allows for the Images to be deleted.   I'm sure there is a better solution, but is working for me, until I understand better whats happening. Below is the traceback.

I also see this in issues #111 and #109.  I have just started using the wiki, added this to my project yesterday.  Is there something else I should be doing to allow users to delete/purge articles and images?  My impression is this is a very broken plugin.  I can't even delete the images from a shell in the condition that the code is in.  

```
Environment:


Request Method: POST
Request URL: http://127.0.0.1:8000/admin/wiki/article/

Django Version: 2.0.4
Python Version: 3.6.5
Installed Applications:
['django.contrib.gis',
 'btcore.apps.BtcoreConfig',
 'crispy_forms',
 'django.contrib.admin',
 'django.contrib.auth',
 'django.contrib.contenttypes',
 'django.contrib.sessions',
 'django.contrib.messages',
 'django.contrib.staticfiles',
 'widget_tweaks',
 'accounts',
 'rest_framework',
 'rest_framework_gis',
 'django_filters',
 'rest_framework.authtoken',
 'btgis',
 'btdocs',
 'posts',
 'markdownx',
 'tinymce',
 'django.contrib.sites.apps.SitesConfig',
 'django.contrib.humanize.apps.HumanizeConfig',
 'django_nyt.apps.DjangoNytConfig',
 'mptt',
 'sekizai',
 'sorl.thumbnail',
 'wiki.apps.WikiConfig',
 'wiki.plugins.attachments.apps.AttachmentsConfig',
 'wiki.plugins.globalhistory.apps.GlobalHistoryConfig',
 'wiki.plugins.notifications.apps.NotificationsConfig',
 'wiki.plugins.help.apps.HelpConfig',
 'wiki.plugins.images.apps.ImagesConfig',
 'wiki.plugins.macros.apps.MacrosConfig',
 'wiki.plugins.links.apps.LinksConfig',
 'btwiki',
 'django_unused_media']
Installed Middleware:
['django.middleware.security.SecurityMiddleware',
 'django.contrib.sessions.middleware.SessionMiddleware',
 'django.middleware.common.CommonMiddleware',
 'django.middleware.csrf.CsrfViewMiddleware',
[...]

PR #1034, opened 2020-03-28
Bugfix/fix delete image

Steckelfisch:
fix for: https://github.com/django-wiki/django-wiki/issues/936

### DIFF
--- a/src/wiki/plugins/images/models.py
+++ b/src/wiki/plugins/images/models.py
@@ -119,13 +119,24 @@ def on_image_revision_delete(instance, *args, **kwargs):
     if not instance.image:
         return
 
-    # Remove image file
-    instance.image.delete(save=False)
-
+    path = None
     try:
         path = instance.image.path.split("/")[:-1]
     except NotImplementedError:
         # This backend storage doesn't implement 'path' so there is no path to delete
+        # return
+        pass
+    except Exception:
+        # in case of other errors
+        # return
+        pass
+    finally:
+        # Remove image file
+        instance.image.delete(save=False)
+
+    if path is None:
+        # This backend storage doesn't implement 'path' so there is no path to delete
+        # or some other error
         return
 
     # Clean up empty directories
--- a/src/wiki/plugins/images/models.py
+++ b/src/wiki/plugins/images/models.py
@@ -121,7 +121,7 @@ def on_image_revision_delete(instance, *args, **kwargs):
 
     # from attachments.models.py (line 166) :
     # Remove file
-    path = instance.image.path.split("/")[:-1]
+    path = os.path.dirname(instance.image.path)
     instance.image.delete(save=False)
 
     '''
--- a/src/wiki/plugins/images/models.py
+++ b/src/wiki/plugins/images/models.py
@@ -121,7 +121,6 @@ def on_image_revision_delete(instance, *args, **kwargs):
 
     path = None
     try:
-        path = instance.image.path.split("/")[:-1]
         path = os.path.dirname(instance.image.path)
     except NotImplementedError:
         # This backend storage doesn't implement 'path' so there is no path to delete
--- a/src/wiki/plugins/images/models.py
+++ b/src/wiki/plugins/images/models.py
@@ -119,17 +119,10 @@ def on_image_revision_delete(instance, *args, **kwargs):
     if not instance.image:
         return
 
-    # from attachments.models.py (line 166) :
-    # Remove file
-    path = os.path.dirname(instance.image.path)
-    instance.image.delete(save=False)
-
-    '''
-    # if backendstorage capabillity checking is required, this should be used.
-    # (in attachments also)
     path = None
     try:
         path = instance.image.path.split("/")[:-1]
+        path = os.path.dirname(instance.image.path)
     except NotImplementedError:
         # This backend storage doesn't implement 'path' so there is no path to delete
         pass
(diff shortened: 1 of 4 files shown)


Steckelfisch:
oh. I closed my branch. That stopped the CI check.

Steckelfisch:
i'll close an re-open to restart CI

Steckelfisch:
restart CI (?)

benjaoming:
Hi @Steckelfisch - I restarted the build on Travis. There's been a lot of API problems between Travis <> Github lately.

src/wiki/plugins/images/models.py
```diff
@@ -119,15 +119,35 @@ def on_image_revision_delete(instance, *args, **kwargs):
     if not instance.image:
         return
 
-    # Remove image file
+    # from attachments.models.py (line 166) :
+    # Remove file
+    path = instance.image.path.split("/")[:-1]
```
benjaoming: This won't work on non-POSIX systems.... I know one such commercial operating system that's quite popular and could be affected :smile: 

```suggestion
    path = os.path.dirname(instance.image.path)
```

src/wiki/plugins/images/models.py
```diff
@@ -119,15 +119,35 @@ def on_image_revision_delete(instance, *args, **kwargs):
...
     instance.image.delete(save=False)
 
+    '''
```
benjaoming: All of this code is commented out?

testproject/testproject/urls.py
```diff
@@ -22,6 +22,18 @@
...
+    try:
+        import debug_toolbar
+        urlpatterns = [
+            re_path('__debug__/', include(debug_toolbar.urls)),
+
+            # For django versions before 2.0:
+            # url(r'^__debug__/', include(debug_toolbar.urls)),
+
+        ] + urlpatterns
+    except Exception as e:
```
benjaoming: We should probably catch the `ImportError` here and not all exception types?

tests/plugins/images/test_views.py
```diff
@@ -242,6 +242,36 @@ def test_purge(self):
...
+                },
+            ),
+            data={"confirm": True},
+        )
+        self.assertRedirects(
+            response, reverse("wiki:images_index", kwargs={"path": ""})
+        )
+        self.assertEqual(models.Image.objects.count(), 0)
+        self.assertIs(os.path.exists(f_path), False)
+
```
benjaoming: This is a really great test, thanks for contributing that!

src/wiki/plugins/images/models.py
Steckelfisch: ok. that same must apply to the 'attachements' plugin too then?
I have no way of testing Windows and/or OSX

src/wiki/plugins/images/models.py
Steckelfisch: it is part of the original code. it seems to check for the support of 'path' on the backend.
(path failed with ValueError when the image was deleted). 

mayebe add the check to the 'attachements' module too while I'm at it then?

Steckelfisch: and add a test with that type of backend? (i do not no which backend applies for that)

testproject/testproject/urls.py
Steckelfisch: Ok.

src/wiki/plugins/images/models.py
Steckelfisch: @superandrew added that part (for S3 support?). 

https://github.com/django-wiki/django-wiki/commit/9c7efff3b9310b6642fd4d7a5fc9456544ced8b8#diff-e54da208fb86cf66506a5f83e74bd916

So i think removing it is not such a good idea after all, and the attachments plugin should be taken care of also.
Ok?

src/wiki/plugins/images/models.py
benjaoming: It's okay to not test on these exotic operating systems as long as we use `os.path.split()` without specifying the path separator or use `os.path.dirname` since it's also platform agnostic.

benjaoming: If you press "Commit suggestion" on my suggestion, it will go strait in, you can also add it manually.

benjaoming: Okay, seeing that the previous code supports backends (such as S3 storage), I'm wondering if you can implement this in a way that doesn't introduce new regressions. Others have opened issues and worked on these storage backends, so we shouldn't break that, even though tests might not cover it.

src/wiki/plugins/images/models.py
benjaoming: Nah, it's not a good idea to remove it, see my comment above.

Your change can probably be adapted into the current pattern?

src/wiki/plugins/images/models.py
```diff
@@ -119,13 +119,24 @@ def on_image_revision_delete(instance, *args, **kwargs):
     if not instance.image:
         return
 
-    # Remove image file
-    instance.image.delete(save=False)
-
+    path = None
     try:
         path = instance.image.path.split("/")[:-1]
+        path = os.path.dirname(instance.image.path)
```
benjaoming: Hi @Steckelfisch -- path seems to be assiged twice here. I think the line above should be deleted?

abohossny (commented):
all

abohossny (approved):

abohossny (commented):
[]()

PR #1037, opened 2020-03-31
Updates to Dutch and Polish translations

benjaoming:
CC: @Steckelfisch - this is how a language would be updated

It sounds like from your message on Transifex that you have to fix this part in Django itself:

```
django/contrib/admin:
msgid "Log in"
msgstr "Aanmelden"
```

Anyways, please ping me back, and I can fetch a new translation from Transifex to include here.


Steckelfisch:
How do i ping you back? (sorry, not that custom to git-hubs workflow)

The use of 'Aanmelden' for "Sign up" in django-wiki is also ambiguos it think. (that works two ways)

Steckelfisch:
The use of 'Aanmelden' for "Sign up" in django-wiki is also ambiguos it think. (that works two ways)

benjaoming:
> How do i ping you back? (sorry, not that custom to git-hubs workflow)

I meant use a mention in this PR for instance -- "@benjaoming" works :)

If you can send me a message here when the translations are done, that'd be great.

Steckelfisch:
Ok @benjaoming .

with the remark that the said ambiguity also applies for django_wiki.
I think 'Aanmelden' should go.

benjaoming:
A bit after opening this PR, I received an email alert from Transifex that Dutch was completed, so apologies for the mis-communication here, I didn't know that you had already completed the translation at the time of flagging an update here :)

Transifex' email notifications seem to perhaps only be sent out once a day.

Issue #1038, opened 2020-04-02
Double-scaped HTML character codes, <, >, & etc


benjaoming:
Tried upgrading Markdown to 3.2 with no luck. Possibly related: https://github.com/Python-Markdown/markdown/issues/802

![image](https://user-images.githubusercontent.com/374612/78242946-3394a800-74e3-11ea-9a1f-3417068d3037.png)

oscarmcm:
Fixed in #1039

PR #1039, opened 2020-04-02
Markdown 3.2 series

benjaoming:
Was looking into #1038 but didn't find a fix.


benjaoming:
Since all tests work, the upgrade seems safe

PR #1061, opened 2020-07-25
Adds Django 3.1 support, drops Django 2.1

atombrella:
https://docs.djangoproject.com/en/3.0/releases/3.0/#id3

Fixes #1056

PR #1063, opened 2020-07-29
Atombrella django 21

benjaoming:
Test to see if Travis CI build will start. It didn't in #1061


benjaoming:
It could seem that our account is stuck in some vacuum between travis-ci.com and travis-ci.org migration. I've tried to add the App as well, but the migration feature is disabled.

PR #1073, opened 2020-09-30
Translation update for Japanese and Brazilian Portuguese

benjaoming:


shuuji3:
Please let me know if there is a problem regarding Japanese translation. :wink:

benjaoming:
@shuuji3 - thanks so much! I did fix a minor issue in Transifex where a translation string's replacement `{variable}` was written as `(variable)`, but otherwise it was perfect! :+1: And I don't really know how this one slipped through Transifex.. I thought that normally it would verify these things.

(this PR is now rebased because of unrelated changes)

PR #1074, opened 2020-09-30
updated requirements.txt

Nikhilgupta18:


benjaoming:
Not sure if this is an intentional PR? This is a reusable app, so we don't pin the requirements.

PR #1086, opened 2020-11-15
Show/suppress "Sign up" link according to ACCOUNT_SIGNUP_ALLOWED setting

iburadempa:
Pretty obvious.

### DIFF
--- a/src/wiki/templates/wiki/base_site.html
+++ b/src/wiki/templates/wiki/base_site.html
@@ -78,7 +78,7 @@
         <li class="nav-item">
         <a class="nav-link" href="{% login_url as wiki_login_url %}{{ wiki_login_url }}">{% trans "Log in" %}</a>
         </li>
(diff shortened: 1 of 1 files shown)


src/wiki/templates/wiki/base_site.html
```diff
@@ -78,7 +78,7 @@
         <li class="nav-item">
         <a class="nav-link" href="{% login_url as wiki_login_url %}{{ wiki_login_url }}">{% trans "Log in" %}</a>
         </li>
-        {% if "ACCOUNT_HANDLING"|wiki_settings %}
+        {% if "ACCOUNT_SIGNUP_ALLOWED"|wiki_settings %}
```
benjaoming, line 81:
```suggestion
        {% if "ACCOUNT_HANDLING"|wiki_settings and "ACCOUNT_SIGNUP_ALLOWED"|wiki_settings %}
```

benjaoming (changes requested):
That makes sense, thanks! We should still check if account handling is enabled.

Issue #1094, opened 2021-01-04
Problems with Editsection Plugin and Markdown in a certain case


OTR:
Trying to move my instance of wiki with existed database from 0.5 version to 0.7 I faced a problem when some of articles don't render. Poking around with article's content I found out that the certain sequence of markdown tags breaks down the web app causing internal server error (500).
Looking at logs and release notes I found out that there is a new plugin came out called "`editsection`". Traceback of crashed app leads to the same plugin. Then I tried to reproduce an error turning on and off that plugin by editing `INSTALLED_APPS`. That is how I came to this test scenario.

**$ python --version**
3.6.9

**$ python ../bin/django-admin.py --version**
3.1.5

**$ uname -a**
Linux laptop 5.4.0-58-generic #64~18.04.1-Ubuntu SMP Wed Dec 9 17:00:52 UTC 2020 i686 i686 i686 GNU/Linux

**$ pip list**
Package             Version                 Location
------------------- ----------------------- -----------------------------------------------------------
appdirs             1.4.4
asgiref             3.3.1
attrs               20.3.0
beautifulsoup4      4.9.3
black               20.8b1
bleach              3.1.5
cfgv                3.2.0
click               7.1.2
coverage            5.3.1
cssselect           1.1.0
dataclasses         0.8
distlib             0.3.1
Django              3.1.5
django-classy-tags  2.0.0
django-functest     1.0.4
django-js-asset     1.2.2
django-mptt         0.11.0
django-nyt          1.1.6
django-sekizai      2.0.0
django-webtest      1.9.7
EasyProcess         0.3
entrypoints         0.3
filelock            3.0.12
flake8              3.7.9
furl                2.1.0
identify            1.5.11
importlib-metadata  3.3.0
importlib-resources 4.1.1
lxml                4.6.2
Markdown            3.2.2
mccabe              0.6.1
more-itertools      8.6.0
mypy-extensions     0.4.3
nodeenv             1.5.0
orderedmultidict    1.0.1
packaging           20.8
pathspec            0.8.1
Pillow              8.1.0
pip                 20.3.3
pkg-resources       0.0.0
pluggy              0.13.1
pre-commit          2.9.3
py                  1.10.0
pycodestyle         2.5.0
pyflakes            2.1.1
pyparsing           2.4.7
pyquery             1.4.3
pytest              5.3.5
pytest-cov          2.10.1
pytest-django       3.10.0
pytest-pythonpath   0.7.3
pytz                2020.5
[...]

PR #1091, opened 2020-12-28
User `.iter()` instead of long deprecated and removed method `.getiterator`

benjaoming:
Fixes #1083 

Thanks @teury for finding the solution!

PR #1098, opened 2021-01-10
Editsection removed method .getchildren(), fix bytes-like object

teury:
2 problems are solved

- [x] Expected string or bytes-like object. 
#1094 
(r"\w+", header)) > (r"\w+", str(header)))

- [x]  Use iter(x) or list(x) instead of x.getchildren() 
#1091
https://docs.python.org/3.9/whatsnew/3.9.html#removed


benjaoming:
Thanks for fixing this @teury 

I added the regression test and verified that it indeed failed on the previous master.

PR #1110, opened 2021-03-14
Add first draft for a ‘what links where’ plugin

Anaphory:
I'm fully aware that this pull request leaves a lot to be desired – not just formally (no tests, no documentation, I tested it only for the py38 I have locally installed), so please – see this as a feature request with a very naive implementation, give me some feedback and don't hesitate to close it if the feeback points out that this should be handled quite differently. I haven't worked with Django or wiki before, so this is a deep dive into the matter hoping that I might be able to use it for a project which would need this functionality.

Issues I'm aware of:
- Updating the database in the render process is *really* bad. Where should I move it? How would I hook into the article just after it's submitted?
- Multiple internal links are double-listed.
- The filtering for ‘should the user be able to read the article’ is missing.
- The `whatlinkswhere/network` pages don't filter for the tree below the parent article yet, they give the whatlinkswhere network for the whole root. (And they don't display a network, which would be the plan at some point, but outside the scope for now.)
- How should non-existent articles be handled? Presumably it would actually be nice to know about them and give them ‘what links here’ pages. What about the ‘what links where’ networks? Currently I use an Article-to-Article model, but with a URL-to-URL-model, I could track non-existent link targets.

How would you address these issues? What are the issues that I, new to the framework, are not even aware of?

Would a table of articles linking to other articles maybe even be something that could have a place in core? It could also help with cleanup after move-operations.


benjaoming:
Hi @Anaphory 

Great PR, this is a very good start to getting a new feature in. Thanks! I think your idea sounds perfect for a plugin: Plugins are exactly what they sound like "pluggable" - they may not fit all purposes, but then we don't need a discussion of their usefulness.

I think having tests is the most important part: If they are not tested, then there is nothing to guarantee that they will continue working, and they can become a maintenance problem for the project.

> Updating the database in the render process is really bad. Where should I move it? How would I hook into the article just after it's submitted?

You can use Django signals! So for instance `post_save`. Try to look around in the codebase, and you will find examples. You can also read about it here: https://docs.djangoproject.com/en/3.1/ref/signals/

> The filtering for ‘should the user be able to read the article’ is missing.

This one is an important component, as we cannot introduce a feature that circumvents the permission system unknowingly to the user.

I don't have any comments right now about the plugin's function - I haven't thought so much about. As you are the one who has the use-case, please do feel empowered to make use of your own experiences and ideas here :+1:

Anaphory:
Some progress.

- [X] Updating links now happens due to ArticleRevision's post_save signal.
- [X] Permission checks have been added
  - [ ] Albeit on the Python level, not through smart database queries
- [X] The whatlinkswhere network now filters by the parent URL
- [ ] Now using URLPath as the referenced model – but that doesn't really make it easier to deal with non-existent articles as link targets, because those don't have URLPaths yet, either.
- [ ] Tests are still missing.

Anaphory:
Now that I have a basic idea of what the API could look like, I guess writing tests is next.

Anaphory:
I have some tests now, for the first batch of promised functionality. @benjaoming (or whoever else is able to do it): Could someone do a code review, please?

Missing is the user accessibility tests. I had hoped the `globalhistory` plugin test would show me how to do those, but it doesn't.
Where can I find a good example of tests that set article/subtree permissions and check that a user account which doesn't have those permissions cannot see something?

benjaoming:
@Anaphory as of #1116, Pull Request builds are now back! Would you be able to pull in the latest `main` branch here? Thanks :pray:

Anaphory:
Oh, great. I pass my links as list to the template, not as query result. And in that one tox setup, Paginator seems to have no wrapper around its object list, so `paginator.object_list.count` is the Python builtin list method for counting the frequency of an object in a list, instead of the attribute that gives the length of the object. With https://docs.djangoproject.com/en/3.1/ref/paginator/#django.core.paginator.Paginator.count maybe the `.count()` call fails and it will fall back to `.__len__()`? I'll try that.

Anaphory:
So now that #1115 is merged, I can get back to this.

It needs review, and an idea how to write permissions tests, before it can be merged. And I think I messed with the testproject database, so I should clean that up.

benjaoming:
> @benjaoming How would you prefer changes to the testproject database?

Ideally, changes are minimal -- but since they whole thing should be replaced with fixtures, I think we can accept that they aren't perfect. I would say that it's okay to spend less time on this part. But if you can keep the changes in a single commit that's best for our git history which everyone will have to have pretty much forever :)

> It needs review, and an idea how to write permissions tests

It could be a good idea to write tests that check for negative permissions, i.e. to check that unpriviledged users cannot see names of Articles that are nested in Articles that they do not have access to. Does that make sense?

Unit test approach:

* Write simple Django functions - i.e. if the plugin lives in `wiki.plugins.whatlinkswhere.utils`, then write functions with a simple interface and add some test cases. Be sure to use those same functions for displaying data in UI.
* Integration tests: You could call a view and ensure that title / links do not show up in the response HTML.

src/wiki/models/urlpath.py
```diff
@@ -281,12 +281,12 @@ def create_urlpath(
         if not site:
             site = Site.objects.get_current()
         article = Article(**article_kwargs)
-        article.add_revision(ArticleRevision(title=title, **revision_kwargs), save=True)
         article.save()
         newpath = cls.objects.create(
             site=site, parent=parent, slug=slug, article=article
         )
         article.add_object_relation(newpath)
+        article.add_revision(ArticleRevision(title=title, **revision_kwargs), save=True)
```
benjaoming, line 289: It would be nice with an inline code-comment. I could guess why it's called in this order and why you changed it, it seems to make sense :+1:

src/wiki/plugins/whatlinkswhere/settings.py
```diff
@@ -0,0 +1 @@
+SLUG = "whatlinkswhere"
```
benjaoming, line 1: Some suggestions for slug:

* `related`
* `related-articles`
* `article-network`
* `link-network`

etc?

benjaoming (commented):
I think that the work is looking really great! Really great structure and doesn't look far from finished? Looking forward to reviewing again, sorry that the time is a bit sparse on my side - but will check in again for sure and help my best to get this in!

What would you think about a name link `LinkNetwork` or `ArticleRelations` or `ArticleGraph` or something in this direction?

Anaphory:
> > It needs review, and an idea how to write permissions tests
> 
> It could be a good idea to write tests that check for negative permissions, i.e. to check that unpriviledged users cannot see names of Articles that are nested in Articles that they do not have access to. Does that make sense?

Yes, of course. I just don't know, technically, how to implement a test with a non-default user. Could you point me to an existing test that does that?

src/wiki/plugins/whatlinkswhere/settings.py
Anaphory: `link-network` sounds okay. `graph` sounds to graphical where a visualization is an option, but shouldn't be the default; and because it includes red links, it's really the network of `link`s, not of articles. `related` is just too generic.

Would you suggest that the whole plugin should then also be called `link_network`? That might reduce confusion.

benjaoming: I think that `link_network` sounds great and can perhaps invite further use cases and solutions to be added to the plugin :+1:

benjaoming: Perhaps using `link-network` (with a dash) for the URL slugs for the sake of convention.

benjaoming:
You can observe https://github.com/django-wiki/django-wiki/blob/main/tests/core/test_accounts.py to see a few ways of creating accounts and logging in as a user.

But looking at the current set of tests, I would like to apologize for the suggestion: Asking for a higher test standard than what is already in place is not cool. I think you should design the tests with primarily coverage in mind, and whatever else that you'd like to add is welcomed but not required - you can see that existing view tests do not look at permissions: https://github.com/django-wiki/django-wiki/blob/99fb076924326d3a8f568cbcc581315343da83e5/tests/core/test_views.py#L53

src/wiki/plugins/whatlinkswhere/settings.py
Anaphory: After thinking about it, I think I'll go for linknetwork to avoid any confusion whether it's a `-` or a `_`.

Anaphory:
I have renamed the plugin to `linknetwork` and given it a permissions test. I think with that it's ready to go from my end.

src/wiki/plugins/linknetwork/templates/wiki/plugins/linknetwork/linknetwork.html
```diff
@@ -0,0 +1,70 @@
+{% extends "wiki/base.html" %}
+{% load wiki_tags i18n humanize %}
+
+{% block wiki_pagetitle %}{% trans "Link Network" %}{% endblock %}
+
+{% block wiki_contents %}
+
+<h1>{% trans "Link Network" %}</h1>
+
+{% spaceless %}
```
benjaoming, line 10: Curious quick comment (can review more thoroughly soon) : Why the `{% spaceless %}`?

Anaphory: Ehm. I think I copied another list template – maybe globalhistory? – and adapted it minimally, because I am still new to Django's template language. I assume this is an artefact of that process.

benjaoming: Ah okay :) It's nice to keep the spaces normally - `{% spaceless %}` is mostly for things where spaces should be removed, often because you have them for readability in the template, but once rendered they're wrong (examples: text emails and things inside `<textarea>`)

src/wiki/plugins/linknetwork/models.py
```diff
@@ -0,0 +1,129 @@
...
+__all__ = ["InternalLink", "store_links"]
+
+
+class InternalLink(models.Model):
+    """This model describes links between articles."""
+
+    from_article = models.ForeignKey(
+        "wiki.Article",
+        on_delete=models.CASCADE,
+        verbose_name=_("from_article"),
```
benjaoming, line 31: The `verbose_name` properties and `help_text` can be used to add a bit more clarity about what this field is, also the `to_article` field. They are a bit hard to grasp :) Would be nice to have it written out.

src/wiki/plugins/linknetwork/views.py
```diff
@@ -0,0 +1,127 @@
...
+        if article.current_revision and article.current_revision.deleted:
+            return redirect("wiki:deleted", article_id=article.id)
+
+        if not article.can_read(request.user):
+            return response_forbidden(request, article, read_denied=True)
+
+        return super().dispatch(request, article, *args, **kwargs)
+
+    def get_queryset(self):
+        # TODO: This filter should use query logic, instead of python filtering, as much as possible.
```
benjaoming, line 62: Maybe there isn't much way around this! But you could try looking at the related manager, maybe `self.article.incoming_links.can_read(user=user)` can be used?

Minor thing related to another comment: What about deleted articles? Should they be filtered out as well? You can filter on something like `self.article.incoming_links.filter(current_revision__deleted=False).can_read(user=user)`.

src/wiki/plugins/linknetwork/views.py
```diff
@@ -0,0 +1,127 @@
...
+    def get_context_data(self, **kwargs):
+        # Apparently a standard hack for (ListView, ArticleMixin) classes
+        kwargs_article = ArticleMixin.get_context_data(self, **kwargs)
+        kwargs_listview = ListView.get_context_data(self, **kwargs)
+        kwargs.update(kwargs_article)
+        kwargs.update(kwargs_listview)
+        return kwargs
+
+
+class GlobalUpdate(View):
```
benjaoming, line 119: What does this view do? It doesn't seem to have any authentication protection or permission checking?

src/wiki/plugins/linknetwork/wiki_plugin.py
```diff
@@ -0,0 +1,31 @@
...
+
+from . import settings
+from . import views
+
+
+class LinkNetworkPlugin(BasePlugin):
+    slug = settings.SLUG
+
+    # TODO: By default, show the 'what links here', which in turn should link to the network.
+    article_tab = (_("What links here"), "fa fa-sitemap")
```
benjaoming, line 14: That's a great icon for it :)

benjaoming (commented):
This is really great work. Please see the comments below.

src/wiki/plugins/linknetwork/views.py
```diff
@@ -0,0 +1,127 @@
...
+            if link.from_article.can_read(self.request.user)
+        ]
+
+    def get_context_data(self, **kwargs):
+        # Apparently a standard hack for (ListView, ArticleMixin) classes
+        kwargs_article = ArticleMixin.get_context_data(self, **kwargs)
+        kwargs_listview = ListView.get_context_data(self, **kwargs)
+        kwargs.update(kwargs_article)
+        kwargs.update(kwargs_listview)
+        return kwargs
```
atombrella, line 75: This is supported from Python 3.6+. You can do it the other places.

```suggestion
        return {**kwargs_article, **kwargs_listview}
```

benjaoming:
Hi @Anaphory - I was wondering if you would be interested in pushing this over the finish line? There are some comments and suggestions, but they are far eclipsed by the effort that you've already made :pray:

Anaphory:
> I was wondering if you would be interested in pushing this over 
> the finish line? There are some comments and suggestions, but 
> they are far eclipsed by the effort that you've already made 
> :pray: 
I intend to – unfortunately I have to prioritise other things at 
the moment.

Issue #1114, opened 2021-03-17
Absolute links can be wiki links – how to parameterize tests?


Anaphory:
One red link test, as well as the markdown extension of the red link plugin (otherwise the test would fail, of course), assume that absolute links are external links. This is not true if the wiki root is the document root.
Where does this assumption come from? Why should it be there? If I imagine a flat wiki, where most articles are immediate children of the wiki root, then nearly all internal links are in fact best phrased as absolute links, aren't they?

I asked nearly the same question, although from a different and slightly more naïve perspective, [in the google group](https://groups.google.com/g/django-wiki/c/ympYo09BYJU).

https://github.com/django-wiki/django-wiki/blob/321c0ad913dc0937350da6420079961d5ac48eaa/tests/plugins/redlinks/test_redlinks.py#L28-L30

Anaphory:
This discrepancy is even more visible in

https://github.com/django-wiki/django-wiki/blob/3843e51ed280f2da6d18bdd30be250916a570ab7/tests/plugins/macros/test_links.py#L7-L14

Clearly, by design, this wiki link should be an internal link, yet the test asserts that this link be styled `wiki-external`.

Anaphory:
How could we parameterize the tests over both the case where the wiki root is the root of the server's paths and where the wiki root is not the root of the server's paths? I think this would be useful, but I don't understand the testing infrastructure well enough.

Anaphory:
There are some URL manipulations in [test_urls.py](https://github.com/django-wiki/django-wiki/blob/master/tests/core/test_urls.py).
Naïvely, I would add a new test class where it matters that essentially derives from the base TestCase I want to modify and ‘mixes in’ (not literally, I guess?) test_urls.ArticleModelReverseMethodTest. Would that be a sensible way to go about this?

benjaoming:
> One red link test, as well as the markdown extension of the red link plugin (otherwise the test would fail, of course), assume that absolute links are external links. This is not true if the wiki root is the document root.
Where does this assumption come from? Why should it be there?

I'm not sure where the assumption comes from. It sounds wrong. Absolute links like `/wiki/article-slug` should some be external.

But you can have django-wiki active at different places in the Django urlconf. So for instance. `/admin` or `/static/` can exist as a link in a wiki article but be treated as an external link. I think that this may have been the intention?

Anaphory:
I think with mixin of the url test urlconfig into the red link tests (part of #1115), this is solved for now.

PR #1115, opened 2021-03-17
Re-structure divide between internal and external links

Anaphory:
Addresses, but doesn't fix completely, isssue #1114 
In particular, there is no parameterization over different wiki roots yet, which would be necessary to make sense of several tests.

### DIFF
--- a/src/wiki/decorators.py
+++ b/src/wiki/decorators.py
@@ -1,15 +1,17 @@
 from functools import wraps
 from urllib.parse import quote as urlquote
 
+from django.http import Http404
 from django.http import HttpResponseForbidden
 from django.http import HttpResponseNotFound
 from django.http import HttpResponseRedirect
-from django.shortcuts import get_object_or_404
 from django.shortcuts import redirect
 from django.template.loader import render_to_string
 from django.urls import reverse
 from wiki.conf import settings
 from wiki.core.exceptions import NoRootURL
+
+from . import models
 
 
 def response_forbidden(request, article, urlpath, read_denied=False):
@@ -34,8 +36,41 @@ def response_forbidden(request, article, urlpath, read_denied=False):
         )
 
 
+def which_article(path=None, article_id=None, **kwargs):
+    # fetch by path
+    if path is not None:
+        urlpath = models.URLPath.get_by_path(path, select_related=True)
+        if urlpath.article:
+            # urlpath is already smart about prefetching items on article
+            # (like current_revision), so we don't have to
+            article = urlpath.article
+        else:
+            # Be robust: Somehow article is gone but urlpath exists...
+            # clean up
+            urlpath.delete()
+            raise models.URLPath.DoesNotExist()
+
+    # fetch by article.id
+    elif article_id is not None:
+        # TODO We should try to grab the article from URLPath so the
+        # caching is good, and fall back to grabbing it from
+        # Article.objects if not
+        article = models.Article.objects.get(id=article_id)
+        try:
+            urlpath = models.URLPath.objects.get(articles__article=article)
+        except (
+            models.URLPath.DoesNotExist,
+            models.URLPath.MultipleObjectsReturned,
+        ):
+            urlpath = None
+
+    else:
+        raise TypeError("You should specify either article_id or path")
+    return article, urlpath
+
+
 # TODO: This decorator is too complex (C901)
-def get_article(  # noqa: max-complexity=23
+def get_article(  # noqa: max-complexity 19
     func=None,
     can_read=True,
     can_write=False,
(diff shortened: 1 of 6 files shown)


benjaoming:
@Anaphory as of #1116, Pull Request builds are now back! Would you be able to pull in the latest `main` branch here? Thanks :pray:

Anaphory:
Hm, some `whatlinkswhere` seems to have snuck in here. It would probably good to get you to review this PR independent of that new plugin.

Anaphory:
As said in #1114, it would be nice to parameterize tests over whether the server root is the wiki root or not. That would allow the different options to be covered – we are missing, in particular, the case where a link is server-internal, but not into the wiki; but we want to keep the case where a link is to the server root and that's part of the wiki, which we lose if we just move the wiki in the path hierarchy of the server.

benjaoming:
> As said in #1114, it would be nice to parameterize tests over whether the server root is the wiki root or not.

Agreed - you are welcome to focus on getting this #1115 in firstly, but if you mix this change into this PR, that's also absolutely okay! 👍 I think your approach in this PR look promising. Can definitely find some time to review this, please LMK when you would like an pair of eyes.

Anaphory:
@benjaoming I think I found a good approach, now let me see how many branches I left uncovered.

Anaphory:
Wohoo!

src/wiki/decorators.py
```diff
@@ -34,8 +36,41 @@ def response_forbidden(request, article, urlpath, read_denied=False):
         )
 
 
+def which_article(path=None, article_id=None, **kwargs):
```
benjaoming, line 39: Since this is nested in the `decorators` module and isn't a decorator, it would be better to mark it internal, for instance renaming it to `_which_article`.

Another thing might be that the name is a bit odd. Perhaps just `_get_article` is good for now, simply suggesting that it's the internal part of the `get_article` decorator?

src/wiki/decorators.py
```diff
@@ -71,68 +106,36 @@ def get_article(  # noqa: max-complexity=23
...
+                )
+                path = "/".join(pathlist[:-1])
+                parent = models.URLPath.get_by_path(path)
+                return HttpResponseRedirect(
+                    reverse("wiki:create", kwargs={"path": parent.path})
+                    + "?slug=%s" % pathlist[-1].lower()
+                )
+            except models.URLPath.DoesNotExist:
+                return HttpResponseNotFound(
+                    render_to_string(
```
benjaoming, line 133: The Codecov message above might be relevant.. is it not reaching because of a missing test or because something is wrong in the implementation? :thinking:

benjaoming (commented):
This looks really good, thank you! I had a few questions -- would be happy to review this soon again and get it in :+1: Really well done PR :+1:

src/wiki/decorators.py
Anaphory: This is the case when the URLPath exists, but the corresponding article is gone, and its parent URLPath does not exist.

Given that each URLPath points to its parents, this sounds like it's an an impossible case (at least in the absence of race conditions, and writing a test to explicitly check whether whether this race condition is handled gracefully sounds a bit out of scope) – it's ‘handled’ in my patch because I took a very close transformation of the existing decorator, and it seemed to handle this case.

src/wiki/decorators.py
Anaphory: The reason I extracted this function is because it is useful in other places – I needed it for understanding where links lead in redlinks and in whatlinkswhere. I would move it out of decorators into urls or util, but I wanted to keep it in a single-file diff for easier comparison for this PR.

src/wiki/plugins/redlinks/mdx/redlinks.py
```diff
@@ -65,19 +44,30 @@ def get_class(self, el):
...
+            return self.external_class
+
+        if target.app_names != ["wiki"]:
+            # Links outside wiki
             return self.external_class
+
         try:
-            URLPath.get_by_path(target)
-        except URLPath.DoesNotExist:
+            article, urlpath = which_article(**target.kwargs)
```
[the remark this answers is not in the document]
Anaphory: - It could be an article link, instead of a link given by URLPath.
 - It could be an URLPath that exists, but has no article (the decorator/now which_article function points that option out)
 - It could be a deleted article, which would still be an existing URLPath
 - etc.

None of these are non-existant URLPaths errors, all of this should be red links.

In any case, it seemed right to use the same internal logic for “what article does this path refer to” in both the `get_article` decorator and any other function that wonders to which article a hyperlink leads (i.e. here for redlinks, and in whatlinkswhere) – that was the whole point of extracting the `which_article` function. Implement it with all its complexity once, and then re-use it levelling differences where necessary.

src/wiki/plugins/redlinks/mdx/redlinks.py
```diff
@@ -65,19 +44,30 @@ def get_class(self, el):
...
         # Ensure that path ends with a slash
         relpath = url.path.rstrip("/") + "/"
-        target = urljoin_internal(self.my_urlpath.path, relpath)
-        if target is None:
-            # Relative path goes outside wiki URL space => external
+        try:
+            target = resolve(urljoin(self.my_urlpath, relpath))
+        except Resolver404:
+            # Broken absolute link
+            return self.external_class
```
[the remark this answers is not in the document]
Anaphory: No, the broken slug should be connected to (at least the option of, resolve doesn't know whether article creation needs permissions) the article creation view, which is a valid view. Or am I wrong? If I'm right, this set of lines should be covered by the redlinks test with server root != wiki root only. I'll try to check.

(That is, until someone managed to hack the wiki so far that it doesn't even resolve wiki article creation urls – and then what are they using `redlinks` for anyway, this might as well be a dead external link.)

src/wiki/decorators.py
benjaoming: I've just noticed that there is also a `response_forbidden` function mis-placed in the `decorators` module, which is also not really a decorator, so the strictness would be a bit over-kill.

src/wiki/decorators.py
benjaoming: Thank you so much for clarifying, I was not able to dwell for long enough to reach this level on enlightenment. But what you say makes absolute sense!

src/wiki/plugins/redlinks/mdx/redlinks.py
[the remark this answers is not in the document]
benjaoming: I can see that you are spot-on here -- there isn't any way that this fix should scope-creep into resolving things in other ways than what is already offered and used throughout the codebase :+1: Thanks for clarifying!

benjaoming (approved):
I had an eager to understand some few things here, sorry that it took so long. The fixes are great, thank you for taking the time!

src/wiki/decorators.py
Anaphory: As I said, this was done with an eye towards transparency. I'm actually in favour of cleaning it up, but making it explicit that it's a cleanup, not part of this restructuring.

Issue #1129, opened 2021-08-02
Refactor fixes made for issue #1099 concerning depreciation of add() in Markdown


iqbalabd:
Fixes were made to remove deprecation warning as  per #1099 but repetitive code can be further refactored as per the comment [here](https://github.com/django-wiki/django-wiki/pull/1125#issuecomment-890386989).

iqbalabd:
The [pull request](https://github.com/django-wiki/django-wiki/pull/1130) for this issue has been merged

Issue #1099, opened 2021-01-11
Deprecations warnings from Markdown


benjaoming:
We have a series of deprecation warnings that should be fixed.

They are easy to reproduce by running this test:

```
# Install test requirements
pip install -e '.[test]'
# Run this test
pytest tests/plugins/editsection/test_editsection.py
```

The deprecation warnings:

```
  /home/user/.virtualenvs/django-wiki/lib/python3.8/site-packages/markdown/core.py:124: DeprecationWarning: The 'md_globals' parameter of 'wiki.plugins.editsection.markdown_extensions.EditSectionExtension.extendMarkdown' is deprecated.
    ext._extendMarkdown(self)

tests/plugins/editsection/test_editsection.py::EditSectionTests::test_broken_content
  /home/user/.virtualenvs/django-wiki/lib/python3.8/site-packages/markdown/inlinepatterns.py:206: DeprecationWarning: Flags not at the start of the expression '^(.*?)(?:(?im)\\[imag' (truncated)
    self.compiled_re = re.compile(r"^(.*?)%s(.*)$" % pattern,

tests/plugins/editsection/test_editsection.py::EditSectionTests::test_broken_content
  /home/user/.virtualenvs/django-wiki/lib/python3.8/site-packages/markdown/inlinepatterns.py:206: DeprecationWarning: Flags not at the start of the expression '^(.*?)(?:(?im)\\[imag' (truncated)
    self.compiled_re = re.compile(r"^(.*?)%s(.*)$" % pattern,

  /media/archive/code/django-wiki/src/wiki/plugins/images/markdown_extensions.py:33: DeprecationWarning: Using the add method to register a processor or pattern is deprecated. Use the `register` method instead.
    md.inlinePatterns.add("dw-images", ImagePattern(IMAGE_RE, md), ">link")
```

iqbalabd:
I have made a pull request [here](https://github.com/django-wiki/django-wiki/pull/1125) for this issue.

benjaoming:
Fixed in #1125 - thanks :pray:

PR #1130, opened 2021-08-02
Refactor code for issue #1129

iqbalabd:
I've tried refactoring the repetitive code and reverted the code before #1099

### DIFF
--- a/src/wiki/core/markdown/__init__.py
+++ b/src/wiki/core/markdown/__init__.py
@@ -54,27 +54,41 @@ def article_markdown(text, article, *args, **kwargs):
 
 
 def add_to_registry(processor, key, value, location):
-    """Utility function to add to Markdown's Registry"""
+    """Utility function to register a key by location to Markdown's registry.
 
-    """ Register a key by location. """
+    Parameters:
+    * `processor`: Markdown Registry instance
+    * `key`: A string used to reference the item.
+    * `value`: The item being registered.
+    * `location`: Where to register the new key
+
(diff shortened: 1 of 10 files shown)


benjaoming (approved):
This is looking good :+1: Perhaps expanding the docstring could be helpful for future developers. And then there was some linting stuff, try to run `tox -e lint` locally to reproduce?

benjaoming:
Thanks for addressing my feedback!!

Can you see this page? https://app.circleci.com/pipelines/github/django-wiki/django-wiki/68/workflows/eac8194b-d2cb-4bb4-b10f-cbc96700ce98/jobs/694 (I'm unsure if there's some issue with other folks having access)

If not, the linting issues are these 2:

image

iqbalabd:
@benjaoming Thank you for looking at this and approving. I can access the CircleCI page.

I've tried to fix the linting issues as much as possible but there is still one more test which is not passing and I have run out of ideas on why this is an issue, so I will leave this as it is.

benjaoming:
That's understandable -- the last failing test is a very marginal slip in over-all test coverage. This can happen when duplicate (but tested) code is written in a shorter way. But it's no problem here.

benjaoming:
The way to fix it would be to add a unit test of `add_to_registry` that tests all the possible input values. But I think it's low priority.

benjaoming:
Thanks for the contribution!

iqbalabd:
Thank you @benjaoming !

PR #1134, opened 2021-08-02
Fix local variable issue in notification

### LINKED ISSUE #1132: Plugins notifications problem
When I tried to make my own plugin for my personal project I come accross notification issue when saving an attachment. By debuging. I look into the code that send the notification and think that there is the following issue in **src/wiki/plugins/notifications/apps.py**
```
def ready(self):

    ...

    for plugin in registry.get_plugins():

        notifications = getattr(plugin, "notifications", [])
        for notification_dict in notifications:

            @disable_signal_for_loaddata
            def plugin_notification(instance, **kwargs):

                ... #here when plugin_notification is called the value of notification_dict is the same that at the end of ready()

            signals.post_save.connect(
                plugin_notification, sender=notification_dict["model"]
            )
```
The receiver is called with the configuration of the last notifications add, that doesn't cause much problem since the notifications of _attachments_ and _images_ are almost the same.

SlyPerdrix:
fix #1132

### DIFF
--- a/src/wiki/plugins/notifications/apps.py
+++ b/src/wiki/plugins/notifications/apps.py
@@ -17,33 +17,37 @@ class NotificationsConfig(AppConfig):
         from wiki.decorators import disable_signal_for_loaddata
         from . import models
 
+
+        def get_receiver(notification_dict):
+            @disable_signal_for_loaddata
+            def plugin_notification(instance, **kwargs):
+                if notification_dict.get("ignore", lambda x: False)(instance):
+                    return
+                if kwargs.get("created", False) == notification_dict.get(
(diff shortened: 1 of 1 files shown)


src/wiki/plugins/notifications/apps.py
```diff
@@ -17,33 +17,37 @@ def ready(self):
         from wiki.decorators import disable_signal_for_loaddata
         from . import models
 
+
```
benjaoming, line 20:
```suggestion
```

benjaoming:
Thanks for addressing the comment and fixing the build :100:

PR #1136, opened 2021-08-08
Fix presave

SlyPerdrix:
I looked at the doc for pre_save signals and there is no indication of a "created" parameter. I made a test to show the issue.
Here is my fix + the test, the fix is base on the pre_save signals of attachements revision.


benjaoming:
You are absolutely right, this is not mentioned in the docs. Wondering if this is a previous behavior or it was always a bug. I really think that this was once used and tested, but it's many many years ago since this logic was developed, so quite likely it has broken because there are no test cases. Thanks for adding this!

PR #1165, opened 2022-02-07
Add support for Django 4.0.

atombrella:


benjaoming (approved):
Thanks so much, great to see that Django 4.0 seems to work without changes :tada:

benjaoming:
I think the error seen can be fixed separately here: https://github.com/django-wiki/django-wiki/pull/1167

tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = {py35,py36,py37}-django{22},{py36,py37,py38,py39}-django{30,31,32},lint
+envlist = {py35,py36,py37}-django{22},{py36,py37,py38,py39}-django{30,31,32},{py38,py39,py310}-django{32,40},lint
```
atombrella, line 2: I think this should be simplified a bit, because the environment for Django 3.2 is tested twice with `py38` and `py39`.

benjaoming:
@atombrella https://github.com/django-wiki/django-wiki/pull/1167 is now merged, perhaps you can rebase onto latest `main` and see how that works?

benjaoming:
We are now remaining with a dependency conflict in the `py310-django40` job which will be fixed once django-nyt is updated in https://github.com/django-wiki/django-nyt/pull/105

```
ERROR: Cannot install Django<4.1 and >=4.0, wiki and wiki[test]==0.7.10 because these package versions have conflicting dependencies.

The conflict is caused by:
    The user requested Django<4.1 and >=4.0
    wiki[test] 0.7.10 depends on Django<4.1 and >=2.1
    django-functest 1.0.3 depends on Django>=1.8
    django-mptt 0.11.0 depends on Django>=1.11
    django-nyt 1.2 depends on django<3.3 and >=2.2

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict
```

benjaoming:
@atombrella - can you have a look at the PR? I have a good feeling that it works since the tests are passing.

README.rst
```diff
@@ -25,7 +25,7 @@ The below table explains which Django versions are supported.
 | Release          | Django         | Upgrade from |
 +==================+================+==============+
 | 0.7.x            | 2.2, 3.0, 3.1, | 0.5 or 0.6   |
-|                  | 3.2            |              |
+|                  | 3.2, 4.0       |              |
```
atombrella, line 28: Shouldn't you create a new row for 0.8.x?

atombrella:
> @atombrella - can you have a look at the PR? I have a good feeling that it works since the tests are passing.

Except for the remark regarding a new release 0.8.x and Django 4.0, I think it's good. Since you pushed to my branch, can you fix this, and then merge?

PR #1174, opened 2022-02-20
Enable automatic dependabot PRs

atombrella:
I think it became an opt-in feature of GitHub.

https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/about-dependabot-version-updates

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,8 @@
+version: 2
+
+updates:
+  - package-ecosystem: "pip"
+    directory: "src/"
+    open-pull-requests-limit: 5
+    schedule:
+      interval: "daily"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,8 @@
+version: 2
+
+updates:
+  - package-ecosystem: "pip"
+    directory: "/"
+    open-pull-requests-limit: 5
+    schedule:
+      interval: "daily"
```
benjaoming, line 8:
```suggestion
      interval: "weekly"
```

What's your advice on the frequency of this? As a maintainer, I can rarely react in 1 day anyways, so I just find it fitting to do this weekly?

atombrella: Once a week is fine. Merged dependency bumps are not really worth much, unless you release them :)

benjaoming: Yes exactly.. but they also often requires to ponder if they are actually worth a release or just noisy :)

PR #1184, opened 2022-04-13
Anonymous wiki through nginx (or other proxy)

unsanded:
If you serve the wiki through ngingx, the remote ip is always empty, so 
i added the HTTP_X_REAL_IP header ass an alternative.


benjaoming:
Thanks for quickly fixing this! I'm including some changes from the latest main branch to make the tests build.

Issue #945, opened 2019-02-05
some characters are replaced by HTML escape sequences in code examples


jenda1:
In code examples some characters are displayed as HTML escape sequences.

This markdown source

    Blablabla...

        some_cmd &&

produces the output

     Blablabla...

       some_cmd &amp;&amp;

This is new issue in v0.5, it works in 0.4

benjaoming:
It's probably due to the new implementation of bleach, would really appreciate some help to solve it :)

jenda1:
I will look at it :-)

- jenda

čt 7. 2. 2019 20:35 odesílatel Benjamin Bach <notifications@github.com>
napsal:

jenda1:
similar issue has been fixed in Python-Markdown/markdown#727
I fix it in the same way.

jenda1:
To maintain own copy of codehilite (and fenced_code) is not necessary, we could use the extensions from the Markdown project. But we need to wait for markdown>=3.1 because this issue.

benjaoming:
@jenda1 Is the issue that Markdown does not properly escape characters like `<` in specific versions? Could you put a function in the `wiki.compat` module? For instance calling it `markdown_escape` and then test for the markdown version and return escaped code in case the python-markdown version is within the range that has this error?

By adding the code in this module, it's easy to discover and remove the behavior in later versions (when it's become redundant).

benjaoming:
Fixed in https://github.com/django-wiki/django-wiki/pull/1194

PR #1194, opened 2022-06-27
Fix double escaping of block code

Benbb96:
I faced same issue as https://github.com/django-wiki/django-wiki/issues/945 and I was wondering why its was not merged so I made a Pull Request inspired by these piece of codes : https://github.com/Python-Markdown/markdown/pull/727 and https://github.com/jenda1/django-wiki/commit/07ea6d3360cd54cbbebc5291d34d58c5f3e8453f

### DIFF
--- a/src/wiki/core/markdown/mdx/codehilite.py
+++ b/src/wiki/core/markdown/mdx/codehilite.py
@@ -76,12 +76,24 @@ class WikiFencedBlockPreprocessor(Preprocessor):
 class HiliteTreeprocessor(Treeprocessor):
     """Hilight source code in code blocks."""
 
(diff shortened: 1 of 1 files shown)


benjaoming:
Thanks this is indeed a confirmed bug and you found a really good reference for the fix :+1:

Benbb96:
Thank you very much for the quick merge and deploy to PyPi !

benjaoming:
You're welcome, was super-glad to see someone finally fix this :tada:

benjaoming:
The bug has been "live" on the demo page for a very long time, but it now shows correctly :) https://demo.django-wiki.org/#wiki-toc-code-blocks

PR #1202, opened 2022-08-09
Optimized ArticleRevision queries and added tests for the same

Dev-Mehta:
Hey folks, this is my first time contributing to this repository and I have noticed that there might be some scope for performance improvement in some areas.

To explore the same, I have written one test for `ArticleRevision` model and added it to [`tests/core/test_models.py`](https://github.com/Dev-Mehta/django-wiki/blob/Dev-Mehta-Optimization-Of-Queries/tests/core/test_models.py#L154-L189).

Can someone check the code and verify whether it is fine, if yes I can optimize queries for all models.

Sample Test Result:
image

### DIFF
--- a/src/wiki/views/article.py
+++ b/src/wiki/views/article.py
@@ -597,8 +597,10 @@ class History(ListView, ArticleMixin):
     paginate_by = 10
 
     def get_queryset(self):
-        return models.ArticleRevision.objects.filter(article=self.article).order_by(
-            "-created"
+        return (
+            models.ArticleRevision.objects.select_related("article")
+            .filter(article=self.article)
+            .order_by("-created")
         )
 
     def get_context_data(self, **kwargs):
--- a/tests/core/test_models.py
+++ b/tests/core/test_models.py
@@ -1,3 +1,5 @@
+from time import time
+
 from django.apps import apps
 from django.contrib.auth import get_user_model
 from django.contrib.sites.models import Site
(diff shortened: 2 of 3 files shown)


tests/core/test_models.py
```diff
@@ -148,3 +150,40 @@ def test_articlerevision_presave_signals(self):
...
+            incr_in_pemr = (
+                (normal_runtime - optimized_runtime) / (optimized_runtime)
+            ) * 100
+
+            times_faster = optimized_runtime / normal_runtime
+            print(f"Reduction in time: {red_in_time:.2f}%")
+            print(f"Increase in performance: {incr_in_pemr:.2f}%")
+            print(f"Times faster: {times_faster:.2f}")
+        except ZeroDivisionError:
+            pass
```
benjaoming, line 189: Thanks for adding tests to show the improvements!

Since this is not really a test, can you remove it or replace it with something that actually tests the expected behavior? A test should have a hypothesis that is verified and fails when it doesn't hold.

src/wiki/views/article.py
```diff
@@ -597,8 +597,10 @@ class History(ListView, ArticleMixin):
     paginate_by = 10
 
     def get_queryset(self):
-        return models.ArticleRevision.objects.filter(article=self.article).order_by(
-            "-created"
+        return (
+            models.ArticleRevision.objects.select_related("article")
+            .filter(article=self.article)
+            .order_by("-created")
         )
```
benjaoming, line 604: This change makes a lot of sense :+1:

benjaoming (changes requested):
This is great! Thanks for taking time to look for good simple performance improvements :+1: 

Can you have a look at a way to get rid of the test though? Or replace it with something meaningful?

Dev-Mehta:
> This is great! Thanks for taking time to look for good simple performance improvements 👍
> 
> Can you have a look at a way to get rid of the test though? Or replace it with something meaningful?

Thanks for your appreciation. I wrote the test just to see the change in performance, as I had no idea where to put the code for that other than tests. I will get rid of test once I see performance change for each model and replace them with meaningful new queries.

Once again, thanks a lot🙏

benjaoming (approved):

benjaoming:
Thanks @Dev-Mehta - please do feel free to open up performance improvement PRs in the same manner, it's really helpful with the benchmarks, too :+1:

PR #1204, opened 2022-08-11
(minor) fix "move page" dropdown for Bootstrap 4

chrisv2:
The dropdown in "move page" looks strange and is hard to use due to lack of margins. Adding `dropdown-item` to the outer `<li>` will fix that.


benjaoming:
Ah yes!

image

benjaoming (approved):
Thanks!

PR #1219, opened 2022-10-24
moving sanitization to save rather than display

sodrew:
First off, great work on this project!

I'm suggesting these changes to move the sanitization to the save of the content rather than at it's display.  The reason for this is because I'm hoping to write some plugins that do more dynamic things (e.g., have javascript, use html tags like button), and I don't want to disable or handicap the sanitization at all, since it is important to have.  So instead, I'm proposing you move the logic for sanitization to save, so that intended html created via plugins don't get stripped out.

### DIFF
--- a/src/wiki/core/markdown/__init__.py
+++ b/src/wiki/core/markdown/__init__.py
@@ -30,26 +30,26 @@ class ArticleMarkdown(markdown.Markdown):
 
     def convert(self, text, *args, **kwargs):
         html = super().convert(text, *args, **kwargs)
-        if settings.MARKDOWN_SANITIZE_HTML:
-            tags = (
-                settings.MARKDOWN_HTML_WHITELIST + plugin_registry.get_html_whitelist()
-            )
+        # if settings.MARKDOWN_SANITIZE_HTML:
+        #     tags = (
+        #         settings.MARKDOWN_HTML_WHITELIST + plugin_registry.get_html_whitelist()
+        #     )
 
-            css_sanitizer = CSSSanitizer(
-                allowed_css_properties=settings.MARKDOWN_HTML_STYLES
-            )
+        #     css_sanitizer = CSSSanitizer(
+        #         allowed_css_properties=settings.MARKDOWN_HTML_STYLES
+        #     )
 
-            attrs = {}
-            attrs.update(settings.MARKDOWN_HTML_ATTRIBUTES)
-            attrs.update(plugin_registry.get_html_attributes().items())
+        #     attrs = {}
+        #     attrs.update(settings.MARKDOWN_HTML_ATTRIBUTES)
+        #     attrs.update(plugin_registry.get_html_attributes().items())
 
-            html = bleach.clean(
-                html,
-                tags=tags,
-                attributes=attrs,
-                css_sanitizer=css_sanitizer,
-                strip=True,
-            )
+        #     html = bleach.clean(
+        #         html,
+        #         tags=tags,
+        #         attributes=attrs,
+        #         css_sanitizer=css_sanitizer,
+        #         strip=True,
+        #     )
         return html
 
 
--- a/src/wiki/core/markdown/__init__.py
+++ b/src/wiki/core/markdown/__init__.py
@@ -1,7 +1,7 @@
 import bleach
 import markdown
-from bleach.css_sanitizer import CSSSanitizer
-from wiki.conf import settings
+# from bleach.css_sanitizer import CSSSanitizer
+# from wiki.conf import settings
 from wiki.core.plugins import registry as plugin_registry
 
 
--- a/src/wiki/core/markdown/__init__.py
+++ b/src/wiki/core/markdown/__init__.py
@@ -1,7 +1,7 @@
-import bleach
+# import bleach
 import markdown
 # from bleach.css_sanitizer import CSSSanitizer
-# from wiki.conf import settings
+from wiki.conf import settings
 from wiki.core.plugins import registry as plugin_registry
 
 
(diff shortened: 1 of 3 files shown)


benjaoming:
Hi @sodrew 

Thanks for having the concern and time to do this!

I'm not sure that I agree with the approach, though. But I don't think that it's a problem, I think you can achieve what you want much better and easier by looking at the rules for sanitization.

Currently, we a allow a set of HTML tags, attributes and CSS classes.

Let's say that you have a plugin that has additional needs.

Why not expand the configuration to allow what you plugin wants? Either your plugin could register its own additional tags or your user could add it to their Django project configuration.

What do you think?

Have a look at the available settings here:

 * MARKDOWN_HTML_WHITELIST
 * MARKDOWN_HTML_ATTRIBUTES
 * MARKDOWN_HTML_STYLES

https://github.com/django-wiki/django-wiki/blob/main/src/wiki/conf/settings.py#L87

sodrew:
Indeed, I saw the rules for sanitization, but by allowing certain rules for myself, they enable any user to also incorporate them into their posts, which are security risks that I would not want to open up.  The onClick attribute may be one of the worst examples of this.  This is why I wouldn't want to handicap the rules for sanitization as part of the view, and rather handle sanitization as part of the save so that you can keep sanitization as strict as possible without limiting what a plugin can output dynamically.

I guess my question to you is why do would you want to keep sanitization in the view?  In most cases for multi-user sites, there will be far more views than edits (saves).  Moving sanitization to the save/edit reduces the resource intensiveness of each view call on the server, while being essentially the same w.r.t. current functionality.  Only a few of your CI tests need to be adjusted for this change.

benjaoming:
Btw the sanitization takes place dynamically on purpose, since rendering a wiki article happens with database contents. It can't be done in a static way.

sodrew:
OK, I am not really following your responses, but perhaps we are communicating past each other.    Here are some final responses to your comments though (I appreciate the dialogue and responsiveness!):

> That would be a problem if you introduced a new global default setting.
If my plugin registers allowed tags and attrs through the current means, it will still open the user to enter those tags and attrs, as long as the plugin is enabled and registered on the site.  i.e., if I enable the onclick attr with my plugin, anyone can put in the onclick attr within their post as long as my plugin is functioning.  So, by leveraging the existing implementation's allowable tags and attrs, I still open the site up to security risks and it is essentially a global default setting.  That said, I think we are getting into the weeds here and are losing focus on the broader picture.  Why sanitize on view/output rather than sanitize before it gets entered into the database?  The django way of doing this has form validity in views to follow the basic premise of santizing before saving to the DB.

> Btw the sanitization takes place dynamically on purpose, since rendering a wiki article happens with database contents. It can't be done in a static way.

The changed approach prevents junk from entering the database in the first place.  So you are losing me on how this changed approach would be static.

And no problem if you don't want to consider changing the approach in the near term, I'll just keep building off my fork.  That said, from a cost/benefit standpoint, I don't understand why you wouldn't want to make this change.  The downsides for the changed implementation are little to none (e.g., changing a few CI tests) and have several tangible benefits (e.g., improved performance, tighter security, and more flexibility for plugin developers).

Regardless, thanks for your efforts on this!

benjaoming:
The way we use `bleach` is to call it as the very last step.

```
# data goes in
input: Article model data in Markdown format
save()

# data is displayed
output: Article model data in Markdown format
render()
markdown stuff
bleach
```

This is what I understand the proposal as:

```
# data goes in
input: Article model Markdown data
Data is rendered to HTML
input is bleached
save()

# data is displayed
output: Article model data in readily rendered Markdown format
render()
```

You can't really do that since everything in the Wiki is based around storing Markdown, diffing Markdown etc.

You might be able to create a safe implementation by having a dynamic placeholder for the HTML elements generated by the plugin. It can't be a static placeholder since an attacker could take advantage of it.

But let's say you introduce a new global setting `WIKI_PLUGIN_HTML_POSTPROCESSORS = []` and it's a list of callables that can work on the finally generated HTML? If you have injected a safe placeholder, then your callable can replace it with the unsafe content here.

sodrew:
Ah, I see the disconnect.  Thanks for laying it out clearly and being persistent in the dialogue.  This is what I proposed through the coding changes:

> # data goes in
> input: Article model data in Markdown format
> bleach on markdown format (not fully rendered html from markdown)
> save()
> 
> # data is displayed
> output: Article model data in Markdown format
> render()
> markdown stuff

I overlooked the fact that rendered markdown may still contain items that need to be bleached, though I couldn't think of any examples that would escape the initial bleaching (I tested some as well).  But attackers are so good at finding those corner cases and based on the wisdom at [SO](https://stackoverflow.com/questions/8409284/is-python-markdown-safe-on-untrusted-input#:~:text=The%20Python%20Markdown%20library%20appears%20to%20be%20safe,version%2C%20to%20set%20safe_mode%2C%20and%20to%20set%20enable_attributes%3DFalse.) I see why you are doing what you are doing (bleaching after fully rendered markdown), since I wrongly assumed markdown in safe mode was already stripping things away...

I'm gonna keep working on my project prototype (that uses DJ) and if it goes past prototype, I'll collaborate with ya to build this enhancement to re-inject content on view, post bleach with some global setting.  As an aside, I already am keeping track of what I'm learning about building a plugin and plan to add some add'l documentation that will prove useful for other future plugin builders.  Thanks again for this project (and your responsiveness)!

benjaoming:
I really appreciate the effort and concern, you definitely have good reasons to need a different setup for your plugin.

I would propose that you keep the order of rendering and sanitizing _as is_. But if you need different options for plugins to interact and for instance inject stuff into a post-processing step for the HTML tree, then that's a really good idea :+1: Anything that can accommodate plugins better is great!

There are also some resources about different approaches for Markdown extensions here:

https://python-markdown.github.io/extensions/api/

I do have a feeling that you will still need the extra support of django-wiki to replace some kind of placeholder _after_ sanitizing.

PR #1228, opened 2022-11-19
Update Bootstrap to v4.6.2 and Popper.js v1.16.1

### LINKED ISSUE #1124: Invalid version of popper.js
For bootstrap 4.4.1 it should be popper.js 1.16.0 ([proof](https://getbootstrap.com/docs/4.4/getting-started/introduction/)), not 2.0.5, which [is used now](https://github.com/django-wiki/django-wiki/blob/main/src/wiki/static/wiki/js/popper.js).

With wrong version I am getting error

```
bootstrap.min.js:formatted:991 Uncaught TypeError: u is not a constructor
    at c.t.show (bootstrap.min.js:formatted:991)
    at c.t.toggle (bootstrap.min.js:formatted:970)
    at HTMLButtonElement.<anonymous> (bootstrap.min.js:formatted:1102)
    at Function.each (jquery-3.4.1.min.js:2)
    at k.fn.init.each (jquery-3.4.1.min.js:2)
    at k.fn.init.c._jQueryInterface [as dropdown] (bootstrap.min.js:formatted:1095)
    at HTMLButtonElement.<anonymous> (bootstrap.min.js:formatted:1186)
    at HTMLDocument.dispatch (jquery-3.4.1.min.js:2)
    at HTMLDocument.v.handle (jquery-3.4.1.min.js:2)
```

and dropdowns on wiki pages don't work.

With correct version all is OK.

oscarmcm:
Fixes #1124 

Also add setting for Django 4.X in order to avoid `(models.W042)` warning.


oscarmcm:
@benjaoming question: is necessary to keep the Bootstrap SCSS version? it the end it doesn't offer more than simplify the specificity rules in the markup which can be achieved using the normal bundled CSS and a correct markup.

oscarmcm:
@benjaoming does this looks ok to be merged?

benjaoming:
> @benjaoming question: is necessary to keep the Bootstrap SCSS version? it the end it doesn't offer more than simplify the specificity rules in the markup which can be achieved using the normal bundled CSS and a correct markup.

We are using the SCSS structures offered by Bootstrap in our own SCSS. If I understand your suggestion correctly, it would make it harder to maintain and customize the SCSS at the cost of having fewer files in the repo. Especially for sites that implement their own Bootstrap-based theme.

If we take a step further into the solution space of large projects with a lot of JS and CSS, they fetch the SCSS source via package management (often NPM). We could do the same and then build our own SCSS on top of Bootstrap SCSS sources, and then finally (:cherries: on the :cake:) we could generate our CSS artifacts dynamically instead of storing them in the repo. But because all of this is annoying for people making smaller contributions, it might not be the best idea.. or maybe some hybrid can be invented.

benjaoming (approved):
LGTM, will look a bit around on demo.django-wiki.org to see if there are regressions from the Bootstrap update :+1: 

I'm gonna fix the linting issue....

PR #1229, opened 2022-11-23
Set the right class for collapse and form error

### LINKED ISSUE #1208: Warning of illegal attachment types are auto collapsed and thus invisible
I am reporting a bug in the attachment plugin.

When you try to upload something illegal (SVG, jpg), it used to (just upgrade from 0.5 to 0.9) show a warning in red text and the Upload a new file section and the section is not collapsed.
Now it still has the warning text but the section is automatically collapsed. This is confusing. Also, the text is in black instead of red (I think it's okay).

After trying to upload SVG:
image
No warning is shown, but when you click Upload a new file, it's just hidden there:
image

Django version: 4.0.7
Python version: 3.10.5
Platform: Windows 10
django-wiki version: 0.9

oscarmcm:
Fixes #1208 

I also added a few settings to the dev config in order to make it more easier for test with the local Django project.

### DIFF
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -18,7 +18,7 @@ Release plan
 -----------------
 
 * Removed direct dependency on ``tinycss2``, inherited from ``bleach`` - due to old versions of pip unable to resolve
-
+* Fixed wrong CSS class for collapse elements in Bootstrap 4 :url-issue:`1208` (Oscar Cortez)
 
 0.9
 ---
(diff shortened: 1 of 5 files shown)


benjaoming (approved):
This is exactly the right fix, using Bootstrap styles as they are supposed to be used :100: 

Thanks also for adding useful dev settings.

Mind adding a changelog entry? :)

oscarmcm:
@benjaoming Thanks!

for the changelog entry do you mean, edit this file? https://github.com/django-wiki/django-wiki/blob/main/docs/release_notes.rst

benjaoming:
@oscarmcm yes, that's the one!

benjaoming:
Yay :100:

PR #1238, opened 2022-12-16
Add Discussion templates

oscarmcm:
After this commit will be applied, we'll stop allowing users to create issues and move everything under the discussions side, and use the new [discussion forms](https://docs.github.com/en/discussions/managing-discussions-for-your-community/syntax-for-discussion-category-forms)


benjaoming:
Aaaaaaaah so the existing one doesn't work because it was called `ISSUE_TEMPLATES` ?

https://github.com/django-wiki/django-wiki/tree/main/.github/ISSUE_TEMPLATES

oscarmcm:
> Aaaaaaaah so the existing one doesn't work because it was called `ISSUE_TEMPLATES` ?
> 
> https://github.com/django-wiki/django-wiki/tree/main/.github/ISSUE_TEMPLATES

Yes! And I was thinking that maybe we should move everything to the discussions and then if its or not a bug create an issue from the discussion (we do now in the opposite way), kinda like this https://github.com/alpinejs/alpine/issues/new/choose

oscarmcm:
And I believe that `Contributing` are quite different to the `Issue Templates`, contributing is where we explain how to help with Django-Wiki, meanwhile the other is like a schema when a user submits an issue per se.

benjaoming:
Agreed re:CONTRIBUTING. The file was written when it was indeed the first customizable feature that GitHub exposed for the "new issue" page.

Sounds really good to start with discussions and promote to issues, I really prefer that workflow, too :+1:

oscarmcm:
Here's a small comment from the Maintainer of Alpine Js explaining why they switched to it 

https://github.com/alpinejs/alpine/discussions/1423

benjaoming:
I agree with the motivation for Alpine JS, although with the exception that I think it's a mistake to remove the issues altogether. They are a valuable metric for a repo, however NOT a very interesting metric if it's full of dreamy features, hypothesis and support questions :) THOSE I would like to have in the discussion forum.

But I still think that "Add Python 3.11 support" is a clear issue :)

What are your thoughts here? As a maintainer, I guess we can agree that a lot of the stuff on the issue list would be much better in discussions... for instance, talk about adding a new Markdown editor.

oscarmcm:
This is the workflow I'm thinking of:

- User go to issue and gets redirected to discussions
- Create a new discussion "Support for Python 3.11"
- Maintainer review the discussion...
  - Do we have support for Python 3.11?
  - What are the artifacts affected by the new version (Regex)
  - More review and or question
- Maintainers now believe that this is a valid issue
- Transform the discussion into a new Issue
- ...Work...
- Close issue

I think this is clear example of the real benefit on using this new workflow. Most of the issues that we have received are "question like" style than a real bug, so I believe that this will make the maintaining and triage process more cleaner and pretty straightforward

oscarmcm:
This seems to be very cool:

https://docs.github.com/en/discussions/managing-discussions-for-your-community/creating-discussion-category-forms

benjaoming:
@oscarmcm I agree with this approach, that's exactly what I was hoping you'd say! Thanks for elaborating it :+1:

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,11 @@
+blank_issues_enabled: false
+contact_links:
+  - name: Bug Report
+    url: https://github.com/django-wiki/django-wiki/discussions/new
```
oscarmcm, line 4: This URL might change once I figured how the discussion forms works in the end.

.github/DISCUSSION_TEMPLATE/bug-report.yml
```diff
@@ -0,0 +1,51 @@
...
+      description: A clear and concise description of what you expected to happen.
+  - type: dropdown
+    attributes:
+      label: Python Version
+      options:
+        - 3.7
+        - 3.8
+        - 3.9
+        - 3.10
+        - 3.11
```
benjaoming, line 32: Oh this one will require maintenance, so I don't like it so much :)

What do you think about a more simple type of field that says "full traceback, error message, log files" and one that says "Versions of django-wiki, Django and Python"?

.github/DISCUSSION_TEMPLATE/bug-report.yml
```diff
@@ -0,0 +1,51 @@
...
+    attributes:
+      label: Django Version
+      options:
+        - 2.2
+        - 3.0
+        - 3.1
+        - 3.2
+        - 4.0
+        - 4.1
+        - 4.2
```
benjaoming, line 45: Same as with Python version field, I don't think we should require this detail of maintenance

.github/DISCUSSION_TEMPLATE/feature-request.yml
```diff
@@ -0,0 +1,32 @@
...
+  - type: markdown
+      attributes:
+        label: Please describe your feature.
+        required: true
+        value: |
+          A clear and concise description of what the problem is. Ex. I'm always frustrated when ...
+  - type: markdown
+    attributes:
+      label: Describe the solution you'd like.
+      description: A clear and concise description of what you want to happen.
```
benjaoming, line 23:
```suggestion
      description: A clear and concise description of what you want to happen. If relevant, add some context about your use case or organization.
```

.github/DISCUSSION_TEMPLATE/feature-request.yml
```diff
@@ -0,0 +1,32 @@
...
+      required: true
+  - type: markdown
+    attributes:
+      label: Describe alternatives you've considered.
+      description: A clear and concise description of any alternative solutions or features you've considered.
```
benjaoming, line 28: I think this is requiring a bit too much?

.github/DISCUSSION_TEMPLATE/feature-request.yml
```diff
@@ -0,0 +1,32 @@
...
+  - type: checkboxes
+    attributes:
+      label: I've already checked other feature proposals
+      required: true
+  - type: dropdown
+    attributes:
+      label: Is your feature request related to a problem?
+      options:
+        - Yes
+        - No
```
benjaoming, line 13: I don't understand this one. I would get really confused about what a "yes" and a "no" actually means here. Maybe just remove it?

benjaoming (approved):
There are some parts where I think there's too many fields to fill in and it'd be nice if it was a bit simpler.

Btw. if you need to merge this into main in order to see it in action and finalize it, please feel free :+1:

PR #1247, opened 2023-01-11
Rework editsection plugin (fixes #1246)

### LINKED ISSUE #1246: Bug: Edit section fails when fenced code block contains comments
### Discussed in https://github.com/django-wiki/django-wiki/discussions/1245

<div type='discussions-op-text'>

<sup>Originally posted by **chrisv2** January  8, 2023</sup>
The following markup will break section editing (due to the `#` in the fenced code block):

````markdown
# Section 1
Section 1 Lorem ipsum dolor sit amet

```python
# hello world
print("hello world")
```

# Section 2
Section 2 Lorem ipsum dolor sit amet
````

# Expected result
When clicking the "edit" link next to "Section 2", the editor should show the following markup:

```markdown
# Section 2
Section 2 Lorem ipsum dolor sit amet
```

# Actual result

The editor shows parts of the fenced code block:

````markdown
# hello world
print("hello world")
```

# Section 2
Section 2 Lorem ipsum dolor sit amet
````

# Notes

Actually it is surprisingly hard to create a proper fix for that, as [it is not possible to get source line numbers from the markdown parser](https://github.com/Python-Markdown/markdown/issues/1235). Currently, the editsection plugin parses the source itself by using regexes, but that doesn't account for fenced code blocks, and it's probably quite difficult to do this with regex alone. IMHO we could do one of the following:

1. loop over markdown source line by line, detecting headers and start/end of fenced code blocks (and possibly other blocks like HTML)
2. create a block level extension and annotate each `h[1-6]` with the source line, then find them in the markup

I'm interested in hacking at this, but I'm not sure which is the best approach. Any thoughts?</div>

### PR DESCRIPTION
**Pull request by chrisv2**
The main problem of editsection is to find the correct location of the section to edit in the source. Previously, that was done with custom, regex-based parsing, but that fails when code blocks contain lines which look like a header (for example, python comments).

This patch uses another approach to find the correct sections. It uses extended versions of the BlockParsers responsible for creating headers, which annotate those headers with the consumed source block. Since we know now which input was responsible for generating the header, we can easily locate it in the markdown source.

### DIFF
--- a/src/wiki/plugins/editsection/wiki_plugin.py
+++ b/src/wiki/plugins/editsection/wiki_plugin.py
@@ -13,7 +13,7 @@ class EditSectionPlugin(BasePlugin):
     urlpatterns = {
         "article": [
             url(
-                r"^(?P<location>[0-9-]+)/header/(?P<header>\w+)/$",
+                r"^header/(?P<header>[\w-]+)/$",
                 views.EditSection.as_view(),
                 name="editsection",
             ),
--- a/src/wiki/core/markdown/__init__.py
+++ b/src/wiki/core/markdown/__init__.py
@@ -13,6 +13,7 @@ class ArticleMarkdown(markdown.Markdown):
         self.article = article
         self.preview = preview
         self.user = user
+        self.source = None
 
     def core_extensions(self):
         """List of core extensions found in the mdx folder"""
@@ -29,6 +30,8 @@ class ArticleMarkdown(markdown.Markdown):
         return extensions
 
     def convert(self, text, *args, **kwargs):
+        # store source in instance, for extensions which might need it
+        self.source = text
         html = super().convert(text, *args, **kwargs)
         if settings.MARKDOWN_SANITIZE_HTML:
             tags = (
--- a/src/wiki/plugins/editsection/markdown_extensions.py
+++ b/src/wiki/plugins/editsection/markdown_extensions.py
@@ -1,19 +1,19 @@
-import re
 import logging
+import re
+from wiki.core.markdown import add_to_registry
+from wiki.plugins.macros.mdx.toc import wiki_slugify
+from xml.etree import ElementTree as etree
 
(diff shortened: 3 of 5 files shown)

### CODE REVIEW

File: src/wiki/plugins/editsection/wiki_plugin.py [hunk 1]
```diff
@@ -13,7 +13,7 @@ class EditSectionPlugin(BasePlugin):
     urlpatterns = {
         "article": [
             url(
-                r"^(?P<location>[0-9-]+)/header/(?P<header>\w+)/$",
+                r"^header/(?P<header>[\w-]+)/$",
```
benjaoming, line 16: Will this work if two section titles are identical?

I.e.

```
# Investigation 1

## Date

# Investigation 2

## Date
```

benjaoming:
I really like the new 2xparsing approach with the PreProcessor added. It seems like a much better way of parsing the document and code has become simpler :+1:

src/wiki/plugins/editsection/wiki_plugin.py
chrisv2: Sure... the TreeProcessor will ensure that each section has a unique ID (if it doesn't have one already, for example from the TOC extension). I will add a test for that to ensure.

chrisv2: oh, good catch :blush:

benjaoming: It sounds like a great great idea to use the same ID for the URL, this will be easy to understand and maintain :+1:

oscarmcm:
Hi @chrisv2 just checking if this PR is ready?

chrisv2:
> Hi @chrisv2 just checking if this PR is ready?

From my perspective it's ready to merge, unless you find some bugs ;)

oscarmcm:
Thanks, I'll take a look and test!

PR #1249, opened 2023-01-15
Pass current revision to Editor.get_widget()/get_admin_widget()

chrisv2:
Sometimes it is useful to have the current revision in the editor widget instance, for example to create preview and upload links, autocomplete, ...

<textarea name="content"
          data-upload-url='/wiki/my-page/_plugin/attachments/...'>

Previously, EditorBase had an instance attribute "instance", but that doesn't make sense (there is only one global instance of Editor) and wasn't ever filled anyway. This patch removes that, renames the "instance" parameter in get_widget()/get_admin_widget() to "revision" (to make clear what this parameter is) and fills it appropriately.

### DIFF
--- a/tests/core/test_editor.py
+++ b/tests/core/test_editor.py
@@ -1,0 +1,79 @@
+import wiki.editors
+from django.forms import Textarea
+from wiki.editors.base import BaseEditor
+
+from ..base import RequireRootArticleMixin
+from ..base import WebTestBase
+from ..base import wiki_override_settings
+
+
+class CustomEditor(BaseEditor):
+    def get_widget(self, revision=None):
+        return Textarea(attrs={"data-revision": revision.pk})
+
+    def get_admin_widget(self, revision=None):
+        return Textarea(attrs={"data-revision": revision.pk})
+
+
+class EditorTest(RequireRootArticleMixin, WebTestBase):
+    def setUp(self):
+        super().setUp()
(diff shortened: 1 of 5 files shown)


tests/core/test_editor.py
```diff
@@ -0,0 +1,79 @@
...
+from ..base import WebTestBase
+from ..base import wiki_override_settings
+
+
+class CustomEditor(BaseEditor):
+    def get_widget(self, revision=None):
+        return Textarea(attrs={"data-revision": revision.pk})
+
+    def get_admin_widget(self, revision=None):
+        return Textarea(attrs={"data-revision": revision.pk})
```
benjaoming, line 15: Very nice editor, requires no JS :)

benjaoming (approved):
Thanks for making the Editor class more useful and adding a nice basic test for using a custom editor :+1: 

I don't have any concerns here, it looks like the old `instance` kwarg is some sort of intention that was never finished and left dangling. "Instance" - of what? :)

oscarmcm (approved):
Thanks for this pull request!

tests/core/test_editor.py
chrisv2: > requires no JS 

Actually true (I'm trying to use CodeMirror, which is ES6) :stuck_out_tongue_winking_eye:

PR #1250, opened 2023-01-18
Upgrade for Sphinx 6 on Read the Docs

benjaoming:
This only works locally, it is pending a fix for https://github.com/readthedocs/readthedocs.org/pull/9654


benjaoming:
This would be more fun in a PR that RTD was building, but I think it won't see this since I enabled PR builds _after_ opening the PR.

Issue #1044, opened 2020-04-11
Editing the demo without a login would be nice 


guettli:
It would be nice, if you could edit some pages in the demo without creating an account.

https://demo.django-wiki.org/

benjaoming:
Oh yeah sorry about that - I think an instruction went missing somewhere. You can log in with admin/admin.

For a better experience, we could:

* [ ] Add a basic instruction to README.rst :)
* [ ] Create a setting for demo mode in `wiki.settings`
* [ ] Customize demo accounts
* [ ] Instruction message box for demo mode w/ login instructions

PR #1260, opened 2023-01-26
Remove sqlite database and use fixtures

oscarmcm:
One step closer to improve the demo

https://github.com/django-wiki/django-wiki/issues/1044#issuecomment-612394655


oscarmcm:
There's a small issue when a new Article is created, seems that there's no relationship for the Article

image

oscarmcm:
Its raised on `django-wiki/src/wiki/models/article.py` at line `120`, in `get_children`

```py
    def get_children(self, max_num=None, user_can_read=None, **kwargs):
        """NB! This generator is expensive, so use it with care!!"""
        cnt = 0
        for obj in self.articleforobject_set.filter(is_mptt=True):
            if user_can_read:
                objects = (
                    obj.content_object.get_children() …
                    .filter(**kwargs)
                    .can_read(user_can_read)
                )
            else:
                objects = obj.content_object.get_children().filter(**kwargs)
            for child in objects.order_by("articles__article__current_revision__title"):
```

I'm thinking if it's related to the drop of the content types in the fixtures.

benjaoming:
This is nice to get done and get rid of this binary blop in the repo. It's not been a giant burden to have the .sqlite file in Git though, many times it made things run faster. But certainly also annoying, and I think _unconventional_ for others to look at.

> I'm thinking if it's related to the drop of the content types in the fixtures.

Can you elaborate this? :pray:

oscarmcm:
>> I'm thinking if it's related to the drop of the content types in the fixtures.

@benjaoming 
> Can you elaborate this? 🙏

I excluded the `ContentTypes` from the export via:

```txt
python manage.py dumpdata --format json --indent 4 --output testproject/fixtures/fixtures.json --exclude auth.permission --exclude sessions.session --exclude admin.logentry --exclude contenttypes.contenttype
```

And then once I load the data in the new sqlite database, this raise the error that the `Settings` doesn't have a `get_children()` attribute, but this doesn't happen on the `prepupalted.db` so I'm not sure if this is happening due to the exclusion of the `ContentTypes` or if there's something when running blank migrations or importing the data.

oscarmcm:
**UPDATE**

Issue #1044 is now the discussion #1263 

*Why?*

There's a lot of different questions and steps to be completed before we can have a decent demo, Discussions allow us to have thread and thus avoid the long citation like format (`>`)

And now, this pull request will be used to track all the required changes in order to have a decent demo.

benjaoming:
It's nice that you are making the issue about improving the demo actionable :+1:

Content-types: I think I have been here before. Django automatically populates contenttypes, right? So when we want our fixtures to contain contenttypes, we either have to

a) wipe all contenttypes and use new ones from the fixture
b) something with "natural keys", not sure if that might work, but worth a shot?
c) Maybe leave some kind of placeholder in the "fixtures" and have a pre-processing step before loading them?

It's all a bit messy. Perhaps this was the original reason why I went with this database... well that, and laziness :)

oscarmcm:
After checking this, and doing a small diff - The issue was related to the content types 

image

oscarmcm:
@benjaoming this can be reviewed now!

benjaoming (approved):
This is really really nice! Glad to see the fixtures are so simple and readable as well :+1:

PR #1253, opened 2023-01-24
Put an upper bound on bleach and comment each and every dependency

benjaoming:
Maybe it's useful to start commenting on why a dependency is there and what expectations we have for the versions that are pinned? What do you think @oscarmcm ? :)

Fixes #1252

Issue #1232, opened 2022-12-12
Python 3.11 support: global flags not at the start of the expression at position 9


k-brahma:
hi, for all the pages in /wiki/ same error raises, "global flags not at the start of the expression at position 9"
do you have any advice?

```python
Internal Server Error: /wiki/manual/

error at /wiki/manual/
global flags not at the start of the expression at position 9

Request Method: GET
Request URL: https://forum.pc5bai.com/wiki/manual/
Django Version: 4.0.8
Python Executable: /home/forumapp/dsite/venv/bin/python3
Python Version: 3.11.0
```

below is the latter part of the stacktrace:
```python
  File "/home/myuser/myapp/venv/lib/python3.11/site-packages/wiki/models/article.py", line 218, in render
    article_markdown(
  File "/home/myuser/myapp/venv/lib/python3.11/site-packages/wiki/core/markdown/__init__.py", line 52, in article_markdown
    md = ArticleMarkdown(article, *args, **kwargs)
  File "/home/myuser/myapp/venv/lib/python3.11/site-packages/wiki/core/markdown/__init__.py", line 11, in __init__
    super().__init__(*args, **kwargs)
  File "/home/myuser/myapp/venv/lib/python3.11/site-packages/markdown/core.py", line 96, in __init__
    self.registerExtensions(extensions=kwargs.get('extensions', []),
  File "/home/myuser/myapp/venv/lib/python3.11/site-packages/markdown/core.py", line 125, in registerExtensions
    ext._extendMarkdown(self)
  File "/home/myuser/myapp/venv/lib/python3.11/site-packages/markdown/extensions/__init__.py", line 77, in _extendMarkdown
    self.extendMarkdown(md)
  File "/home/myuser/myapp/venv/lib/python3.11/site-packages/wiki/plugins/images/markdown_extensions.py", line 36, in extendMarkdown
    md.inlinePatterns, "dw-images", ImagePattern(IMAGE_RE, md), ">link"
  File "/home/myuser/myapp/venv/lib/python3.11/site-packages/markdown/inlinepatterns.py", line 209, in __init__
    self.compiled_re = re.compile(r"^(.*?)%s(.*)$" % pattern,
  File "/usr/local/lib/python3.11/re/__init__.py", line 227, in compile
    return _compile(pattern, flags)
  File "/usr/local/lib/python3.11/re/__init__.py", line 294, in _compile
    p = _compiler.compile(pattern, flags)
[...]

PR #1264, opened 2023-01-30
Add support for the latest bleach version

oscarmcm:
This is a continuation of #1253, an improved version of #1255, and fixes:

- #1259 
- #1257 
- #1251

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,7 +40,7 @@ dependencies = [
     # Django version support is added one minor release at a time
     "Django>=2.1,<4.2",
     # bleach has been pretty stable, hence the loose pin
-    "bleach[css]>=5,<7",
+    "bleach[css]>=6,<7",
     # Pillow is used very little and has never broken
     "Pillow",
     # django-nyt is maintained by django-wiki maintainers, so we can
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,7 +40,7 @@ dependencies = [
     # Django version support is added one minor release at a time
     "Django>=2.1,<4.2",
     # bleach has been pretty stable, hence the loose pin
-    "bleach[css]>=5,<6",
+    "bleach[css]>=5,<7",
     # Pillow is used very little and has never broken
     "Pillow",
     # django-nyt is maintained by django-wiki maintainers, so we can
@@ -123,7 +123,7 @@ test = "pytest {args}"
 
 [tool.hatch.envs.test.overrides]
 matrix.django.dependencies = [
-    { value = "django~={matrix:django}.0" },
+    { value = "django~={matrix:django}" },
 ]
 
 [tool.hatch.envs.test]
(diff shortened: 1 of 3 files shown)


benjaoming (approved):
These look like good changes. Wondering if they work with some old version of Bleach.

oscarmcm:
> These look like good changes. Wondering if they work with some old version of Bleach.

I dunno, This is a very breaking change... unless we add middleware to transform old Django wiki settings (list) into a frozenset; but from what I know this wouldn't work out of the box.

benjaoming:
Maybe we should push the lower bound on bleach. Our tests suggest that it's not a problem wrt. Django and Python versions.

The only way that it can cause problems is with other dependencies for users of django-wiki. But I think the chances are narrow, and those might as well consider upgrading Bleach.

pyproject.toml
```diff
@@ -40,7 +40,7 @@ dependencies = [
     # Django version support is added one minor release at a time
     "Django>=2.1,<4.2",
     # bleach has been pretty stable, hence the loose pin
-    "bleach[css]>=5,<6",
+    "bleach[css]>=5,<7",
```
benjaoming, line 43:
```suggestion
    "bleach[css]>=6,<7",
```

pyproject.toml
```diff
@@ -194,7 +194,7 @@ pull = [
 [tool.hatch.envs.docs]
 dependencies = [
     "Django>=3.1.13",
-    "bleach>=3.3.0,<5.1",
+    "bleach>=3.3.0,<7",
```
benjaoming, line 197:
```suggestion
    "bleach>=6,<7",
```

benjaoming:
Hmm, the demo currently looks like this. And we haven't changed anything to Bootstrap. I'm wondering if this is the new Bleach version removing stuff that is intended to be included:

image

However, in the case of "locked", that text isn't supposed to be visible at all.

oscarmcm:
@benjaoming related to the image, I just changed the image tag to use an external source instead of the internal, and thus there's no way for that image to align with plain markdown.

benjaoming:
@oscarmcm roger that, hmm... it looks kind of bad for the demo.. but unrelated to this PR... maybe we can get an image for the fixtures somehow, so it'll look better. Perhaps a management command to setup the fixtures? I've had good experiences in other projects of creating a management command for fixtures to deal with complex stuff.

miguelgr:
Hi @oscarmcm , will these changes be available in a patch version for v9.x ? or will be included in [0.10](https://pypi.org/project/wiki/0.10b1/) 
If I am not wrong, currently by `pip installing wiki` the default version of wiki is v9 which doesn't handle the TypeError for the bleach.ALLOWED_TAGS.
Thanks in advance.

benjaoming:
@miguelgr I think we can release this in a 0.10b2 - would you mind giving some feedback on it if we put up that release?

miguelgr:
I did try `pip install --pre wiki` and I got the site up and running even with the fixes for 3.11 included in #1232, also I guess for the 0.10 release.

benjaoming:
@miguelgr the 0.10b1 version didn't contain this PR AFAICT.

miguelgr:
It did fix the issue that 0.9 has with bleach 6.
Where `wikipedia` is a just created django project with the wiki installed.

image

PR #1267, opened 2023-02-01
Add settings and middleware for demo

oscarmcm:
Final step to complete #1263 - related to https://github.com/django-wiki/django-wiki/issues/1044#issuecomment-612394655


testproject/testproject/settings/demo.py
```diff
@@ -0,0 +1,10 @@
+from .base import *  # noqa
+
+DEBUG = False
+
+INTERNAL_IPS = []
```
oscarmcm, line 5: @benjaoming Not sure if here we'll leave a fixed value or read it from the environment var.

testproject/testproject/fixtures/fixtures.json
```diff
@@ -155,8 +155,8 @@
       "model": "sites.site",
       "pk": 1,
       "fields": {
-          "domain": "example.com",
-          "name": "example.com"
+          "domain": "django-wiki.org",
```
benjaoming, line 158:
```suggestion
          "domain": "demo.django-wiki.org",
```

testproject/testproject/settings/demo.py
benjaoming: These parts are injected by the deployment script so they can be ignored. Unless you know a good default that's needed for local deployments of the demo :+1:

benjaoming (approved):

oscarmcm:
@benjaoming I believe this is ready to go!

testproject/testproject/fixtures/fixtures.json
```diff
@@ -119,7 +119,7 @@
       "model": "auth.user",
       "pk": 1,
       "fields": {
-          "password": "pbkdf2_sha256$390000$IICRDKwC5zBSkzbCfpkUpx$xxVdvVYU2BkEXRbmoyDBcy2paLXAGrkt4COJslZBOgU=",
+          "password": "",
```
benjaoming, line 122: Does this make the password "unusable"?

oscarmcm: yups!

oscarmcm:
@benjaoming I've just added support in the fixtures for the image plugin, so there's no real need now to use a custom management command.

image

Related to the "locked" word, seems to be related to this:
https://github.com/django-wiki/django-wiki/blob/main/src/wiki/templates/wiki/article.html#L24

More than that, this should be ready to merge, lmk.

benjaoming:
Deployed and working :+1: https://demo.django-wiki.org/

benjaoming:
Something up with the admin user, though.. it doesn't seem to have any admin rights:

image

oscarmcm:
I just left the Admin user to be able to add/edit new pages. Not sure what else we must be covering.

benjaoming:
I couldn't get that to work, `admin` doesn't seem to own `use-this-space` so cannot add new pages nor edit anything existing stuff.

oscarmcm:
ohhh I see, there's an small issue in the fixtures

https://github.com/django-wiki/django-wiki/blob/main/testproject/testproject/fixtures/fixtures.json#L297

The article is owned by `djangowiki` super user, changing that to `2` should fix it.

Issue #1276, opened 2023-04-10
Release final 0.10 with Django 4.2 support


benjaoming:
Issue title kind of says it :) Would be nice to do this soon :+1:

oscarmcm:
Quick update here:

We need to also update `Django-nyt` to `Django 4.2` or we'll end with dependencies conflict.

See this https://github.com/django-wiki/django-wiki/pull/1275#issuecomment-1502116588

oscarmcm:
I think we can release `0.10` as Circle CI is passing on the latest update

https://app.circleci.com/pipelines/github/django-wiki/django-wiki/385/workflows/793d7dba-4c15-4012-9d3e-696edf677836

oscarmcm:
@benjaoming do you want me to create a new release or you'll do it?

benjaoming:
@oscarmcm that sounds great, please do!! I usually do this to prepare a release:

1. Check translations (there aren't any)
2. Open PR with the final changes (version bump + release note upgrade)

Wen the repo is ready, I merge it and locally do this:

1. `git checkout main`
6. `git pull`
7. `git tag -s releases/0.10`
8. `git push --tags`
9. `hatch build`
10. `twine upload -s` (or maybe you have hatch operational for this one!)

Please feel free to do how you find is a good way!! Btw I was just reading the hatch documentation and it seems it has nice commands for bumping versions - I didn't know that. If there's anything you'd like to tweak, feel free :+1:

benjaoming:
Edit! There's an error in the `tx pull` command, I think the config file doesn't work with new client versions...

benjaoming:
The issue with the transifex-client is that they finally switched off the API support for the old client. So we'll have to migrate to the new client.

I'm not so happy that it's a binary file you're supposed to download blindly and execute. But they have a reasonable Docker alternative:

https://github.com/transifex/cli#running-from-docker-beta

I'm having some promising results adding this command...

oscarmcm:
Released!

https://pypi.org/project/wiki/

benjaoming:
SOLID!

PR #1284, opened 2023-05-10
Update languages + update to use the new Transifex client in Docker

benjaoming:
Couldn't pull language updates anymore and had to update the Transifex client. It's not on PyPi anymore but some binary blob.

See: https://github.com/django-wiki/django-wiki/issues/1276#issuecomment-1542794054

Does the pyproject.toml change make sense @oscarmcm ? :blush:


oscarmcm (approved):
Thanks for getting into this, the changes looks good to me!

PR #1294, opened 2023-06-12
Fix xframe_options_sameorigin bug in MergeView

liuxiawei:
In class MergeView, add @method_decorator(xframe_options_sameorigin) in line 913 to avoid bug


benjaoming (approved):
Thanks :100:

PR #1296, opened 2023-06-16
Grant more permission to the users

### LINKED ISSUE #1290: Grant more permissions in demo
## Expected Behavior

We need to use the wiki to show and test django-wiki, which involves having admin rights.

## Current Behavior

Current demo site doesn't grant any useful permissions to the `admin:admin` user. So it's impossible edit and add articles :/

## Context (Environment)

https://demo.django-wiki.org/

oscarmcm:
Fixes #1290


benjaoming:
It works :tada:

PR #1298, opened 2023-07-27
Use django.shortcuts.render for 500/404 handler view in testproject

benjaoming:


oscarmcm:
Cool fix! but more cool to finally see `codecov` running after the fixes.

PR #1299, opened 2023-07-31
Editsection plugin: Relax path regex (fixing NoReverseMatch)

chrisv2:
# Problem

The following, valid markdown causes a NoReverseMatch:

```markdown
# Title 1 {#some_id_with.dot}
```

This is due to editsection's url regex not allowing dots and other, non-whitespace characters - we cannot reverse `some_id_with.dot`, even though it is a valid ID as per the HTML5 spec.
    
# Solution
    
Relax the url regex to allow any non-whitespace character, as HTML5 does.


benjaoming (approved):
@chrisv2 This looks good to me :+1: 

Had to lookup the `\S` switch:

> It matches up any character that is not a part of s.

Can you fix the linting? Just need to install and run `pre-commit --from-ref main --to-ref HEAD` (IIRC)

benjaoming:
Thanks for finding and fixing this :+1:

chrisv2:
`pre-commit run --all-files` seems to do the trick. Fixed.

Issue #1042, opened 2020-04-09
Improve ToC plugin to enable feature parity between it an the Article List plugin


lucianoratamero:
One of the biggest issues I'm currently having is that we're using django-wiki as a knowledge base, and since we have really big articles with small sections, our ToC is huge, and there is no way to customize it (we currently rely on the python-markdown extension).

The solution for that would be for the ToC plugin to work just like the Article List plugin, enabling us to have a 'depth' kwarg for better selection of which levels of headers are going to be presented.

lucianoratamero:
I'll probably be working on this myself, but any pointers/contributions would be welcome!

benjaoming:
That sounds like a really great contribution, maybe you can patch up the `[toc]` macro in `wiki.plugins.macros.mdx.macro` to support `depth:123` just like `[article_list]` ?

PR is definitely welcome.

PR #1304, opened 2023-09-26
Update TOC macro to enable feature parity with Article List

orcephrye:
This is a rough draft attempt to resolve Issue: https://github.com/django-wiki/django-wiki/issues/1042

This is inspired by the existing code here: https://github.com/django-wiki/django-wiki/blob/918b354aabf23ae7c1612e9ac746d7c7905c137b/src/wiki/plugins/macros/mdx/toc.py#L16

It caches keyword arguments parsed by the MacroPattern class to a class variable in WikiTreeProcessorClass. Then the 'run' method maps these keyword arguments to the object's attributes before running 'super().run(doc)'. Afterward, it takes a tmp dictionary with the attribute's original values sets, then back and resets the cache that was set by MacroPattern.

I have attempted to follow the instructions for contributing. I used hatch test and lint before submitting, and I added several new tests for TOC. 

**NOTE**: The original tests for TOC didn't test TOC within django-wiki but instead simply tested that the Markdown TOC extension still worked. While this is nice to make sure that this 3rd party dependency maintains this feature as new versions are supported, this doesn't do a good feature test. I may have adjusted the test_toc.py file in a way that is not standard but hopefully will do a better job of testing with all the extensions registered.

Problems to consider: 
What if an exception happens? Perhaps the super().run(doc) should be within a try block with a finally undoing variables. Thoughts?

### DIFF
--- a/src/wiki/plugins/macros/mdx/toc.py
+++ b/src/wiki/plugins/macros/mdx/toc.py
@@ -17,23 +17,14 @@ def process_toc_depth(toc_depth):
 
 def process_bool_value(bool_val, org_val):
     if type(bool_val) is str:
-        if bool_val.lower() == "false":
+        if bool_val.lower() == "false" or bool_val == "0":
             return False
-        elif bool_val.lower() == "true":
+        elif bool_val.lower() == "true" or bool_val == "1":
             return True
         else:
             return org_val
-    elif type(bool_val) is int:
-        if bool_val == 1:
-            return True
-        elif bool_val == 0:
-            return False
-        else:
-            return org_val
-    elif type(bool_val) is bool:
-        return bool_val
     else:
-        return org_val
+        return process_bool_value(str(bool_val), org_val)
 
 
 def process_value(org_val, new_val):
--- a/src/wiki/plugins/macros/mdx/toc.py
+++ b/src/wiki/plugins/macros/mdx/toc.py
@@ -7,12 +7,12 @@ HEADER_ID_PREFIX = "wiki-toc-"
 
 
 def process_toc_depth(toc_depth):
-    if isinstance(toc_depth, str) and '-' in toc_depth:
-        toc_top, toc_bottom = [int(x) for x in toc_depth.split('-')]
+    if isinstance(toc_depth, str) and "-" in toc_depth:
+        toc_top, toc_bottom = [int(x) for x in toc_depth.split("-")]
     else:
         toc_top = 1
         toc_bottom = int(toc_depth)
-    return {'toc_top': toc_top, 'toc_bottom': toc_bottom}
+    return {"toc_top": toc_top, "toc_bottom": toc_bottom}
 
 
 def process_value(org_val, new_val):
@@ -31,9 +31,9 @@ def wiki_slugify(*args, **kwargs):
 
 
 class WikiTreeProcessorClass(TocTreeprocessor):
-    CACHED_KWARGS = dict()
+    CACHED_KWARGS = dict()  # Used to cache arguments parsed by the MacroPattern
+    # Used to map the keyword arguments to the Class Objects attribute name.
     TOC_CONFIG_VALUES = {
-        "marker": "marker",
         "title": "title",
         "baselevel": "base_level",
         "separator": "sep",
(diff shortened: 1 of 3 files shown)


oscarmcm (changes requested):
Think you can improve a bit the test suite for this?

https://github.com/django-wiki/django-wiki/pull/1304/files#diff-bc2ec08479758bcaa23a175d2b50e2c3393482edec61c2101961f6fd2d07b047R24

oscarmcm:
> What if an exception happens? Perhaps the super().run(doc) should be within a try block with a finally undoing variables.

Yeah I think that might be the best way to handle that

oscarmcm:
Nice work with this, having this parity is definitely a good idea.

src/wiki/plugins/macros/mdx/toc.py
```diff
@@ -6,16 +8,92 @@
...
+        toc_bottom = int(toc_depth)
+    return {"toc_top": toc_top, "toc_bottom": toc_bottom}
+
+
+def process_value(org_val, new_val):
+    try:
+        if type(new_val) is str:
+            new_val = new_val.lstrip("'").rstrip("'")
+            if type(org_val) is not str:
+                new_val = json.loads(new_val.lower())
```
benjaoming, line 25: What is going on here? Maybe I'm forgetting something, I haven't been around this codebase for a long time. But I don't understand how JSON comes into the picture :)

src/wiki/plugins/macros/mdx/toc.py
```diff
@@ -6,16 +8,92 @@
...
+                            WikiTreeProcessorClass.TOC_CONFIG_VALUES[k]
+                        ] = getattr(self, WikiTreeProcessorClass.TOC_CONFIG_VALUES[k])
+                        setattr(
+                            self,
+                            WikiTreeProcessorClass.TOC_CONFIG_VALUES[k],
+                            process_value(
+                                tmp_kwargs[WikiTreeProcessorClass.TOC_CONFIG_VALUES[k]],
+                                v,
+                            ),
+                        )
```
benjaoming, line 88: I'm not so concerned with the indentation levels, I guess some people would point that out as an issue.

But it'd be nice with some inline comments on the inner-most parts to add a narrative and make it more readable? Would you mind adding that?

src/wiki/plugins/macros/mdx/toc.py
benjaoming: I'm a little concerned here that it'd be possible to have `json.loads` called with data from an attacker.

benjaoming (commented):
This looks like a really great addition to the TOC extension, making it far more capable + adding the user documentation that was needed :+1: Great job on testing, too!

src/wiki/plugins/macros/mdx/toc.py
orcephrye: Derp! You are totally right. I didn't think of that. 
The reason was I was thinking about the future. Markdown 3.4.4 TOC extension supports an argument 'toc_class' which can take a list of classes to provide the div containing the TOC. I wanted to make sure I could handle all data types. It was a little extra weight and now that you mention it a bad idea.

Instead, I should just assume for now that the value is one of the following, str, int, bool. And act accordingly.

src/wiki/plugins/macros/mdx/toc.py
orcephrye: I will attempt to make this look cleaner while respecting lint settings.

benjaoming: I don't wanna stress this change with "stylistic" changes, but just adding some helpful comments could make it easier to read :+1:

src/wiki/plugins/macros/mdx/toc.py
benjaoming: Thanks for aligning with the Markdown TOC extension as well :100:

src/wiki/plugins/macros/mdx/toc.py
orcephrye: When Django-Wiki supports the latest version of Python-Markdown this will need to be re-evaluated because there is the possibility of taking in arguments that are structured like a list. 

I think I will do a simple check to see if the string starts with [ and ends with ] and then use some regex to part it into a list. However, at this time, that is sorta out of the scope of this PR. 

I would like to point out though, that I originally took on the Issue: https://github.com/django-wiki/django-wiki/issues/1042 because at the time, I didn't realize that Python-Markdown 3.4 wasn't supported. And I wanted to play around with adjusting the class of the div that wraps the TOC. 

So... hopefully in the future!?

oscarmcm:
This looks good! the only final suggestion I have is that codecov is now triggering an alert bcz it decreased from 79.14% to 75%

orcephrye:
Yes, I am changing things in order to help deal with CodeCov. I should have made a branch and then squashed the commits.

oscarmcm (approved):
Thanks for the hard work @orcephrye - This looks amazing!

Issue #1197, opened 2022-07-08
Support for pymdown-extensions


benjaoming:
The default of django-wiki should be to depend on pymdown-extensions and have out-of-the-box CSS support for a set of plugins.

https://facelessuser.github.io/pymdown-extensions/

orcephrye:
I would also love to see this. But there are a lot of extensions, and some of them seem like a challenge to add. 

For example: https://facelessuser.github.io/pymdown-extensions/extensions/snippets/

How would one go about implementing that? Add custom MDs to a static directory? How would that be managed? Or perhaps have a wrapper around the extension to pull down raw MD from a slug?

benjaoming:
@orcephrye Agreed! I think that snippets cannot be supported, especially since it says "Not Meant for User Facing Sites" in its docs :exclamation: 

Actually, the support of the extensions should probably be added 1-by-1. So if someone wants to add support, they could focus on adding it for just the extension that they need.

The philosophy of django-wiki is also to leverage Django models to store and organize knowledge/data. It's like what Wagtail went and did, too. This might mean that support would have to be customized for the extensions. An example would be: https://facelessuser.github.io/pymdown-extensions/extensions/tasklist/ - adding and managing task lists is something I would find significantly more useful if it was stored in a model.

orcephrye:
Agreed I thought about using a model somehow for both tasklist and tables. Not all tables but some may be better suited if the date wasn't in a md file. I'm part because I have some very big tables. 

Well maybe I will look at implementation of one of the extensions. Any suggestions for a first easy one?

benjaoming:
> Well maybe I will look at implementation of one of the extensions. Any suggestions for a first easy one?

I can try to share how I would do this -- but only since you ask :) This will sound a lot like all talk, no action from my side :) So this is just a suggestion... it's my first take on how to approach it:

If you start by introducing the simplest form of extension, I would think it's easier to focus on merely getting a nice structure for the plugins. This rules out the sort of extensions that contains structural data (the kind we want to store in a Django model).

I think that it could make sense to create a top-level package called `wiki.plugins.pymdown` (for common stuff) and have `wiki.plugins.pymdown.<extension_name>` for each extension so individual extensions are enabled through `INSTALLED_APPS`.

I think that the `blocks.*` extensions look extremely useful. They don't contain any data structures that are useful in a database (sure, we could mirror tabs in the database.. but seems useless).

[...]

PR #1316, opened 2023-10-20
Pymdown blocks support

orcephrye:
This is for Issue: #1197 

I need help! 

There are several things that are not complete about this PR. However, it is working and it is a start and I figured I could use this as a way to get help.

1) I am not a CSS person so there are questions around how to get these different blocks to look correct. Particularly with detail/summary tags. I also wanted to use font awesome icons but I couldn't get that to work for some reason. 

2) Currently, the sidebar for the markdown editor only supports macros and seems to exist only within the macros plugin. But it makes sense that PyMdown extensions should show up in that sidebar right? However, I am unsure as to how to proceed with that. 

There may be more things to discuss but this is a starting place.


orcephrye:
Also I don't know why these tests are failing. Running >hatch run test:all had all passing. I will take another look. But these tests have been problematic as of late.

orcephrye:
> Also I don't know why these tests are failing. Running >hatch run test:all had all passing. I will take another look. But these tests have been problematic as of late.

I fixed it. It was a caused bunch some merges that screwed up the pyproject.toml file. The version I had no longer had the pymdown-extension dependency but the package was still installed in my test environments. So it was working for me but failing here.

oscarmcm:
Thanks for the PR @orcephrye I'll take a look on this.

Im just worried by the codecov decrease.

orcephrye:
> Thanks for the PR @orcephrye I'll take a look on this.
> 
> Im just worried by the codecov decrease.

I am extremely confused by this and thus a little worried too. So the 72.72% is from the diff not the whole project.  It looks... sorta like it is complaining that I am not testing the apps.py file? But the apps.py is just a simple little configuration. I can add a 'test' for it but it feels pointless and stupid.

oscarmcm:
yeah, I think the same too.

Let me see what I can do, maybe tell `codecov` to not read that file will help us

oscarmcm:
Thanks for the help with this @orcephrye 🎉

orcephrye:
> Thanks for the help with this @orcephrye 🎉

Hey... I felt like this PR wasn't finished. Sorry for any confusion. Perhaps I should have just linked to a branch and kept talking about Blocks in the Issue instead. But I do not feel that this is a finished solution. It works. If the plugins are enabled then the PyMarkdown blocks (minus Tab) will work. 

However, I have questions about the styling/CSS for the blocks. IE: Does the color fit? Should Icons for the different types be a good idea? How can I use the icons that are already setup in scss? I tried and couldn't figure out how to use them.

Also, I think we should add documentation to the Marco sidebar so that someone who has this plugin enabled can see how to take advantage of it. The problem is that currently that is located within the marcos plugin. Not sure the best way to go about adding to that.

benjaoming:
Good that you mention it @orcephrye - I had the impression that it wasn't finished because of your comment in the PR description, and I originally wanted to have tested it out manually to understand layout issues.

But it's always nice to get things merged :) You can use the "draft" state for the PRs to ensure that there's full clarity about the state of the work. I sometimes also add "WIP" to the PR title to make sure it's really really clear :joy:

benjaoming:
But now harm done here, seems the additional improvements can come in follow-up PRs :+1:

PR #1320, opened 2023-10-26
Update settings.py to allow align attributes in td tags

yengip:
One line change in wiki/conf/settings.py to add "align" as a permitted attribute for td tags.


benjaoming (approved):
Looking good! We'll want to add a little changelog entry, but this is great :+1:

oscarmcm (approved):

PR #1325, opened 2023-11-09
Updates simplify testproject settings (and fix them)

benjaoming:
I've had issues running the testproject locally. It turns out the root cause was `SESSION_COOKIE_DOMAIN = ".localhost"`, however incidentally I stumbled upon that `SESSION_COOKIE_SECURE` can be eliminated as well and be left to `SESSION_COOKIE_SECURE=True` (as defined in `base.py`).

I don't have an explanation for this, I assume something has changed in either Django or browser behavior.

Would you mind testing if this works on your setup, too @oscarmcm ?


oscarmcm:
I added those settings as I had the same trouble running the demo app

See here

https://github.com/django-wiki/django-wiki/commit/714cd034ead97f870984564a6271861fe0ab0e18

benjaoming:
@oscarmcm I understand that you added them, I'm just wondering if something has changed and they're not needed? I don't know what might have changed, but everything works fine for me.

I wrote my observation here: https://forum.djangoproject.com/t/why-are-cookie-secure-settings-defaulted-to-false/1133/4

benjaoming:
Do you have issues running this PR branch?

oscarmcm:
It doesn't let me log-in.

The form succeeds but I believe the project is not setting the session cookie.

benjaoming:
> The form succeeds but I believe the project is not setting the session cookie.

Roger! I want us to have a setting that works for everyone.

For me, it works with `SESSION_COOKIE_SECURE=False` as well, and I don't have to define `SESSION_COOKIE_DOMAIN` -- would that work for you?

oscarmcm:
Leaving just `SESSION_COOKIE_SECURE=False` works for me!

I'm gonna approve this PR then.

oscarmcm (approved):

benjaoming:
Great, will make the change :+1:

benjaoming:
What browser are you using btw?

oscarmcm:
> What browser are you using btw?

I'm usually on Safari

PR #1327, opened 2023-12-02
S3 storage fix - attachments

dev176:
Fixing an error "This backend doesn't support absolute paths." (Attachments, S3 storage)


benjaoming (commented):
The code itself looks good. But maybe from the same source that it was copied (images plugin), it would be possible to adapt some of the tests verifying that attachments are deleted correctly?

See: https://github.com/django-wiki/django-wiki/pull/1034/files

PR #1329, opened 2023-12-05
Combine CodeCov uploads in all test steps

benjaoming:
By uploading the report at each step, we will have the full coverage report. This can include various types of specific support for different Django/Python versions that we are testing against.

I think that this is the way to do it, inputs most welcome!


oscarmcm (approved):

oscarmcm:
Thanks for this!

PR #1334, opened 2024-01-28
Add Pymdown plugin documentation sidebar for markdown editor

orcephrye:
As per the conversation in: https://github.com/django-wiki/django-wiki/issues/1197 

This is a follow-up to previous PRs that added PyMDown plugin. 

It adds a sidebar in the markdown editor just like for the Macros plugin. Is shows documentation for how to use the PyMDown extensions supported in the PyMDown plugin. 

This hopefully concludes #1197 for now.


src/wiki/plugins/pymdown/settings.py
```diff
@@ -27,3 +31,78 @@ def update_whitelist():
...
+    {
+        "short_description": _("Admonition"),
+        "help_text": _("Insert an Admonition html tag"),
+        "example_code": """
+/// admonition | Some title
+    type: warning
+
+Some content
+///
+More information: https://facelessuser.github.io/pymdown-extensions/extensions/blocks/plugins/admonition/
```
benjaoming, line 46: I found it confusing that this text is included in the example. At first, I thought that it was part of the syntax, then realized that it's a helpful note.

Can it be moved to the `help_text` part? (same for the other ones)

benjaoming (commented):
GREAT WORK! :tada:  I was just testing it out, and it's super nice to have these extensions included.

I was wondering if you can either move or remove the help text that's added in the code example?

src/wiki/plugins/pymdown/settings.py
orcephrye: Done. I added it to the help_text part. Thanks for catching that. Doesn't belong in the example code section.

benjaoming:
@orcephrye just a note that you've force-pushed out changes that I made to your PR. Ideally, if you see this situation, you should pull in the changes. Only force-push when you know what you're doing :)

I had it stored locally, so I've been able to cherry-pick my change now.

src/wiki/plugins/pymdown/settings.py
benjaoming: Great! Thanks! :)

benjaoming (approved):
Great work, thanks for addressing the feedback! Tested locally and seems to be a great addition :+1:

orcephrye:
> @orcephrye just a note that you've force-pushed out changes that I made to your PR. Ideally, if you see this situation, you should pull in the changes. Only force-push when you know what you're doing :)
> 
> I had it stored locally, so I've been able to cherry-pick my change now.

I didn't know you made a change. I always force push when I squash commits. I made a commit to add the changes requested, Squashed, force pushed. I should have fetched/pulled and seen an update first sorry. I wouldn't have squashed my commits together then.

benjaoming:
Ah yes, I used to follow that workflow, too - but it's fine to leave this out, since GitHub has "Squash and merge" now. So we can do this when concluding the PR... or the maintainer can choose if they want to merge or squash merge.

PR #1355, opened 2024-03-22
Updates to release plan

benjaoming:


benjaoming (commented):
@oscarmcm this was purely cosmetic :) But let's talk about the roadmap any time!

oscarmcm:
@benjaoming so we're only missing a major FE rewrite with bootstrap?

benjaoming:
@oscarmcm I mean, we can put anything in the roadmap - so perhaps it's a good idea to indicate that other releases should be planned as well?

oscarmcm:
I like that idea, I think for 0.12 bootstrap upgrade is a good milestone + any fixes needed from 0.11

PR #1313, opened 2023-10-05
Python markdown 34 support

orcephrye:
As per:
Discussion:  https://github.com/django-wiki/django-wiki/discussions/1312
Issue: https://github.com/django-wiki/django-wiki/issues/1197
Issue: https://github.com/django-wiki/django-wiki/pull/1274

This updates the Wiki requirements to Markdown 3.4.4

All tests pass on Python 3.8. I need to configure Hatch to see 3.9 and 3.10 (I am skipping 3.7 as it is deprecated). 

I also have yet to run the tests with Selenium, but I have experience in that just haven't yet. 

Appreciate any help and advice to see this PR through.

PR #1359, opened 2024-04-04
Fix attribute error

henrikhorluck:
Maybe introduced in https://github.com/django-wiki/django-wiki/pull/1313

This has been causing errors in our django-wiki installation after upgrading to 0.11.


benjaoming (approved):
Thanks so much for finding and fixing this. It looks like the right fix and diagnose.

I'll let it simmer until the weekend to see if I can drop in a test case that catches it, too.

In any case, will release a patched version ASAP.

oscarmcm (approved):

benjaoming:
Here we go! :)

benjaoming:
It's been released! Thanks @henrikhorluck :+1:

Issue #1322, opened 2023-11-03
Reorganize SCSS sources


benjaoming:
It's hard to contribute to styling because the current layout of SCSS files doesn't reveal which files are original Bootstrap files (that should not be touched) and which are custom files (where changes should happen)

oscarmcm:
Now, there's no real need to use a bunch of SCSS files, or are we going to continue supporting this?

https://getbootstrap.com/docs/4.6/getting-started/theming/

benjaoming:
I would think that we still need to have the source Bootstrap files that we are building with? But we can split our own custom overrides from those so that everything from Bootstrap is kept separately and can easily be updated.

benjaoming:
I think the layout that they suggest looks good:

```
your-project/
├── scss
│   └── custom.scss
└── bootstrap/
    ├── js
    └── scss
```

oscarmcm:
> I would think that we still need to have the source Bootstrap files

Yes, we still need the imports like this:

```sass
// Custom.scss
// Option A: Include all of Bootstrap

// Include any default variable overrides here (though functions won't be available)

@import "../node_modules/bootstrap/scss/bootstrap";

// Then add additional custom code here
```

But I think that this will no longer be required to [have all of this](https://github.com/django-wiki/django-wiki/tree/main/src/wiki/static/wiki/bootstrap/scss) unless you still want to continue using the current structure.

benjaoming:
I don't think we should have any Node stack in the project for _this_ reason. That'd have to happen because of something more important.

I have so far benefited from having the Bootstrap source files in the repo because I often read the code.

benjaoming:
> But I think that this will no longer be required to [have all of this](https://github.com/django-wiki/django-wiki/tree/main/src/wiki/static/wiki/bootstrap/scss) unless you still want to continue using the current structure.

I want to continue having the Bootstrap source files but separate them from the files that contain overrides/additions.

I'd want to avoid any Node installation of Bootstrap for as long as possible, but certainly consider it if there were other things from Node :+1:

benjaoming:
I want to also remove the CSS artifacts from the repo. I find that hatch has so-called "hooks" to ensure that something happens during the build process. We can use those hooks to build the SCSS.

I know that we currently have the `assets` command that we can run, but I'd like somehow that the build and release process doesn't come with a risk of un-synced assets by baking the assets building into the wheel/sdist building.

[...]

PR #1363, opened 2024-05-21
Add support for build scss files on hatch build

oscarmcm:
Initial work for:
  - #1322

---

SCSS changes:

```diff
diff --git a/src/wiki/static/wiki/bootstrap/scss/wiki/wiki.scss b/src/wiki/static/wiki/bootstrap/scss/wiki/wiki.scss
index b12569d0..6d86709c 100644
--- a/src/wiki/static/wiki/bootstrap/scss/wiki/wiki.scss
+++ b/src/wiki/static/wiki/bootstrap/scss/wiki/wiki.scss
@@ -11,21 +11,23 @@ $danger:              #721817;
 {
     background: $info;
 }
 
 @mixin wiki-page-header{
     @extend .pb-2;
     @extend .mt-4;
     @extend .mb-2;
     @extend .border-bottom;
 }
-
+.some-very-long-class {
+  color: yellow;
+}
 .wiki-article .thumbnail
 {
     clear: both;
     margin-bottom: 15px;
     margin-left: 10px;
     margin-right: 10px;
     text-align: center;
 }
```

Output:
image


oscarmcm:
I moved away from pure `SCSS` implementation and used `PySCSS` since those provide pre-built wheels for Linux and maxOS

oscarmcm:
@benjaoming I'll need help figuring out how to add libsassc inside of the read the docs CI

benjaoming:
Awesome work!

This will make it much easier to build releases without maintaining the CSS artifact in git.

Maybe we can argue this way? "libsass" and "pysass" are build dependencies. So we need to define the build dependencies explicitly (not sure what is the "hatch way" of doing that)

For now, I've added it as a docs dependency... but maybe there's a better way?

benjaoming:
Ah, this reminds me of when setup.py had a hack to specify that it needed some requirements to be installed BEFORE installing with `setup_requires`.

benjaoming:
I think the current error is something about the build environment being isolated from the environment that the documentation dependencies are installed into. So I guess that whole approach won't work :confounded:

oscarmcm:
Yeah, that's the thing! 

Our approach works for CircleCI since we can have packages installed on each matrix environment. And since this is not allowed I dont think it will work with the RTD CI environment.

I dont think that marking this as a build dependency is a good idea, I really like the approach to have concerns separate and dont block Python development, just because a contributor doesn't have properly configured a SCSS installation.

Do you know if there's a way to check if the RDT CI has an environment variable that we can read and conditionally run the command?

benjaoming:
Yes, you can define environment variables on RTD :+1:

benjaoming:
> I really like the approach to have concerns separate and dont block Python development, just because a contributor doesn't have properly configured a SCSS installation.

Agreed!

oscarmcm:
@benjaoming this should be good to go :)

pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["hatchling"]
+requires = ["hatchling", "hatch-build-scripts"]
```
benjaoming, line 2: Great! So I guess that it's not possible to list libsass here?

I've clicked "watching" on the hatch-build-scripts repo since it seems like it's a small project: https://github.com/rmorshea/hatch-build-scripts

pyproject.toml
```diff
@@ -78,6 +78,17 @@ Documentation = "https://django-wiki.readthedocs.io/en/latest/"
 Tracker = "https://github.com/django-wiki/django-wiki/issues"
 Source = "https://github.com/django-wiki/django-wiki"
 
+[[tool.hatch.build.hooks.build-scripts.scripts]]
+out_dir = "src/wiki/static/wiki/bootstrap/css/"
+commands = [
+    "[[ -z $READTHEDOCS_VIRTUALENV_PATH ]] && pysassc --style compressed src/wiki/static/wiki/bootstrap/scss/wiki/wiki-bootstrap.scss src/wiki/static/wiki/bootstrap/css/wiki-bootstrap.min.css || true",
```
benjaoming, line 84: Great solution :+1:

benjaoming (approved):

pyproject.toml
oscarmcm: That's a good observation!!

Based on the PEP documentation, we can place it there (https://peps.python.org/pep-0518/) I could explore that just in case, the only thing I'm not OK is that using this solution will make the lib sass required for everything, since for each environment this is build on dev mode.

benjaoming: Not a bad idea. I don't have any strong opinions on this one, if the libsass dependency isn't really problematic and pysassc is fast to run, then it's okay for me. But we can easily just wait and see if the new setup causes any problems, it also seems fine as it is :+1:

PR #1375, opened 2024-08-21
Hotfix before releasing: wiki-bootstrap.min.css missing in dist files

benjaoming:
@oscarmcm there was a misconfiguration in this rather complex area... I noticed that the .css file was missing in the .whl.

I _think_ (not totally sure) that the right way is to say that our `pysass` command is enough to generate the artifacts in their right location. Additional management of artifacts isn't necessary.

I'm gonna go ahead and merge this now in order to get the release out, but I think it does the right thing :)


oscarmcm (approved):

benjaoming:
I think it gets moved to the wrong location, then afterwards the original is cleaned up.

benjaoming:
Maybe, just maybe, there is something else wrong... let me check again!

PR #1376, opened 2024-08-21
Fix the hotfix

benjaoming:
No release yet, but I was just about to do one, that's why I wanted #1375 with a hotfix.. which of course got rushed.

And sorry, this is gonna get confusing @oscarmcm 

I actually had some issues where I was missing pysassc, but after installing it, I still get the artifact in the wrong location.

(I had fixed that wrongly in the hotfix, too)

But it seems to produce everything fine now.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -92,7 +92,7 @@ else:
             '--style',
             'compressed',
             'src/wiki/static/wiki/bootstrap/scss/wiki/wiki-bootstrap.scss',
-            'src/wiki/static/wiki/bootstrap/scss/wiki/wiki-bootstrap.min.css'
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -92,7 +92,7 @@ else:
             '--style',
             'compressed',
             'src/wiki/static/wiki/bootstrap/scss/wiki/wiki-bootstrap.scss',
-            'src/wiki/static/wiki/bootstrap/scss/wiki/wiki-bootstrap.min.css'
+            'src/wiki/static/wiki/bootstrap/css/wiki-bootstrap.min.css'
```
benjaoming, line 95: This isn't something you had done wrong, @oscarmcm - I was debugging why files ended up in the wrong location, and while doing that I had removed and wrongly restored this line.

PR #1380, opened 2024-09-16
Bootstrap 5, SCSS compile w/django-compressor and django-libsass

rsp2k:
- Upgraded to bootstrap 5. Moved SCSS from hatch buildscript to django-compressor and django-libsass to make SCSS development easier.

### DIFF
--- a/src/wiki/templates/wiki/base_site.html
+++ b/src/wiki/templates/wiki/base_site.html
@@ -36,7 +36,7 @@
         <path d="M8 12a4 4 0 1 0 0-8 4 4 0 0 0 0 8zM8 0a.5.5 0 0 1 .5.5v2a.5.5 0 0 1-1 0v-2A.5.5 0 0 1 8 0zm0 13a.5.5 0 0 1 .5.5v2a.5.5 0 0 1-1 0v-2A.5.5 0 0 1 8 13zm8-5a.5.5 0 0 1-.5.5h-2a.5.5 0 0 1 0-1h2a.5.5 0 0 1 .5.5zM3 8a.5.5 0 0 1-.5.5h-2a.5.5 0 0 1 0-1h2A.5.5 0 0 1 3 8zm10.657-5.657a.5.5 0 0 1 0 .707l-1.414 1.415a.5.5 0 1 1-.707-.708l1.414-1.414a.5.5 0 0 1 .707 0zm-9.193 9.193a.5.5 0 0 1 0 .707L3.05 13.657a.5.5 0 0 1-.707-.707l1.414-1.414a.5.5 0 0 1 .707 0zm9.193 2.121a.5.5 0 0 1-.707 0l-1.414-1.414a.5.5 0 0 1 .707-.707l1.414 1.414a.5.5 0 0 1 0 .707zM4.464 4.465a.5.5 0 0 1-.707 0L2.343 3.05a.5.5 0 1 1 .707-.707l1.414 1.414a.5.5 0 0 1 0 .708z"></path>
       </symbol>
     </svg>
-<div class="dropdown position-fixed bottom-0 end-0 mb-3 me-3 bd-mode-toggle">
+<div style="z-index:10001;" class="dropdown position-fixed bottom-0 end-0 mb-3 me-3 bd-mode-toggle rounded-3 bg-body-tertiary">
   <button class="btn btn-bd-primary py-2 dropdown-toggle d-flex align-items-center" id="bd-theme" type="button" aria-expanded="false" data-bs-toggle="dropdown" aria-label="Toggle theme (auto)">
     <svg class="bi my-1 theme-icon-active" width="1em" height="1em"><use href="#circle-half"></use></svg>
     <span class="visually-hidden" id="bd-theme-text">Toggle theme</span>
@@ -129,7 +129,7 @@
                 <a class="dropdown-item" href="{% url 'wiki:signup' %}">{% trans "Sign up" %}</a>
               </li>
             {% endif %}
-          {% endif %}           
+          {% endif %}
             <li><hr class="dropdown-divider"></li>
           {% block wiki_about %}
             <li><a class="dropdown-item" href="https://github.com/django-wiki/django-wiki" target="_blank">{% trans "About" %}</a></li>


oscarmcm:
this looks very interesting 👀

benjaoming:
@rsp2k you've done a lot of work! Would you like to share some overview/context/analysis? And are you available to address review feedback?

rsp2k:
> @rsp2k you've done a lot of work! Would you like to share some overview/context/analysis? And are you available to address review feedback?

Thank you for the compliment and looking at my patch! I learned a lot from working on this project. I'm here to address review feedback. Major issue is that I can't figure out how to "collectstatic" before tests run? Would really appreciate help here.

Here's where this patch came from... I wanted to add a wiki to my various django projects. Found yours, saw it didn't have bootstrap5 support (yet). Read the history in github issues. Determined a couple of things. Hatch is pretty great (never used it before). This project's SCSS build pipeline could use a little help from django-libsass/django-compressor. This allows you to compile SCSS during the request cycle (cached in non-DEBUG mode). It gives you line numbers and reasonable error messages.

NB, django-libsass replaces the hatch build script that was building SCSS. No more needing to trigger that script.


Once I had bootstrap 5.3.3 integrated, I went through each template and made the necessary tweaks for the new bootstrap version.

Note, I saw that the logo might be up for grabs. I put together a simple SVG (it flashes the cursor once, but not when rendered inside the page???). Just a thought, I really liked the \/\/ in the old logo, makes me wanna grab a computer shopper, lookup a BBS type someAT fire up an good ANSI terminal and fire up some AT commands :) The SVG is really simple to edit if you want to play around with it.

I moved the search bar out of the header because it makes it easier to integrate the app into an existing project. In my case, I only want the Wiki search bar in wiki pages.

Minor tweaks to the existing SCSS files, I'm sure someone with more SCSS skills than I can make some improvements there and possibly add more classes so that they can be referenced.

benjaoming:
Sounds really great! Happy to adopt a new logo, have no feelings for the old one :joy: 

> Major issue is that I can't figure out how to "collectstatic" before tests run? Would really appreciate help here.

Can you explain the need here? (it's not so common that `collectstatic` is part of test setups / test cases)

### Regarding django-compressor

I've used it in a lot of projects. I never had issues with it. It's great. I think we've never even discussed using it. Perhaps it comes from some idea that it was harder to install for users, but I think that the need for system libraries has been removed years ago.

What are your thoughts here @oscarmcm ?

oscarmcm:
@rsp2k great work here! and thank you for taking care of all of this changes, this has been in our backlog for a loooong time, so seeing the light after the dark is mindful.

Related to the compressor or not compressor, I think the most complicated part is bundling the end user with something very opinionated, that's the reason why we prefer having the .css build on the release process, allowing the end user to use it out of the box, or if they want just use whatever package and compile it by hand.

I personally prefer this idea to let the user decide instead of lock-in with a package.

rsp2k:
I understand what @oscarmcm and @benjaoming are saying, but I disagree. I will argue if that were true, you are already "locked-in" (I prefer invested) in every other library this project uses. I'm trying to undetstand how this is different.

With the change I proposed, it required setting up SCSS properly and doesn't require any system packages, (checkout the django-libsass) page. Both django-compressor and libsass install with one line in the requirements, two stanzas in the settings and two lines around the SCSS include in the base template.

My intent was to make minimal changes and provide the most delightful development experience possible. 

With this patch, all a person has to do is hatch create, start the project edit SCSS files "on-the-fly".  When it comes to production deployment, it offers features for compressing/pre-compiling assets.

Have you had a fired up the patch and see the changes yet?

oscarmcm:
@rsp2k agree! Thank you for taking the time to answer. If there's anything I could be helpful, please let me know. More than happy to get this ball out of the park.

benjaoming:
I already used django-wiki with django-compressor. For me, it's quite a natural choice when needing to customize the CSS. I wouldn't want to overwrite CSS or append rules. I would want to import SCSS sources, hence I use django-compressor.

> start the project edit SCSS files "on-the-fly"

One thought is that in order to "unlock" the achievement for users, we need to document the preferred way to extend the SCSS sources from the package. That can come in a separate change, but you're welcome to share your experiences with this @rsp2k :+1: 

 @rsp2k are you able to address test failures? There's a lot of `compressor.exceptions.UncompressableFileError`.

benjaoming:
> Related to the compressor or not compressor, I think the most complicated part is bundling the end user with something very opinionated, that's the reason why we prefer having the .css build on the release process

I tend to agree with this, but I don't think django-compressor is controversial in the Django world. It would be more if someone has preferences to build the SCSS with other external tools.

I'm not sure about the new setup though, if it's nicer -- because now the user has to run `django manage.py compress` in order to ensure that SCSS is built?

oscarmcm:
I think we can continue shipping the build like we did before, to avoid end user this step. 

So we win in both areas and removes the pysassc build.

rsp2k:
The user doesn't have to do anything, that's the beauty of it. If you want,
for production builds, you can precompile assets.

----------
***@***.***
208.991.6341
*.:||:.* Supported Systems <http://supportedsystems.com/>

benjaoming:
@rsp2k would you be able to show in this PR how you would pre-build assets for the release with django-compressor?

rsp2k:
That's just it, nothing to do! staticfiles
<https://docs.djangoproject.com/en/5.1/ref/contrib/staticfiles/>,
django-compressor <https://github.com/django-compressor/django-compressor>
and django-libsass <https://github.com/torchbox/django-libsass> FTW!

I think it's great to be able to leverage other great project's work. It
seems like that's a goal of this project, to be one of those handy apps I
can pip install into my existing project and now I have a wiki!

It took me way too long to learn how to integrate SCSS and deal with static
assets in my django projects. I admit, my front-end development experience
is not my strong suit. Hopefully others can see this particular build
pipeline and chime in.

Basically, if you follow the django-compressor and django-libsass
instructions (installing 2 modules, adding a few lines to settings), and
setup your SCSS includes properly, you're good to go. Bootstrap docs
provide information on how they recommend using their SCSS source.

It just works. Edit your SCSS files and refresh your browser, if you're in
DEBUG mode, SCSS compile errors will be displayed.


----------
***@***.***
208.991.6341
*.:||:.* Supported Systems <http://supportedsystems.com/>

benjaoming:
@rsp2k I think you miss the point here - we all agree that django-compressor is great for someone who is writing SCSS. The question is: How do we ensure that pre-built CSS is bundled in the package (.whl)?

rsp2k:
> @rsp2k I think you miss the point here - we all agree that django-compressor is great for someone who is writing SCSS. The question is: How do we ensure that pre-built CSS is bundled in the package (.whl)?

Have you checked my branch and tried it?

I might be misunderstanding your question. There is no prebuilt CSS, it's generated on the fly (or, optionally by management command).  You're question is very similar to what I asked previously, how do I ensure the collectstatic process runs before the tests? The next thing to figure out is what is required for packaging.

Due to my lack of knowledge, I can't figure out why the tests are not running (at all), yet the app seems to work fine with manual testing. I'm not sure how to fix it and could use some help.

Summary of the SCSS "pipeline" I've setup: 
By default, when a request is made, django-compressor launches libsass as a JIT compiler and returns CSS in the response.

This means during development, when you edit SCSS and refresh the browser, you're looking at the code. Not only does it handle compiling, but also filenaming based on hash of contents to prevent cache issues. If the SCSS has a runtime issue, the errors are displayed in the console. (You're IDE probably catches them before they happen, mileage may vary).

If you are running a high volume site and don't want the overhead of JIT SCSS compile happening per request, you can run a management command (django-compressor) to compile them and then serve the pre-compiled version, using staticfiles to distribute to CDN, or whatever. 

This might be what you meant? Generate these files in the whl. I've not done much packaging for public projects so I don't know what best practice is here. It seems treating them like any other static file is the best route and let the user decide from there. It's a safe assumption that most of the time, the project is being used for development and hobby/small projects. If you're running a high volume site, you'll probably have knowledge and "unique-process".

Setting up the project to by-default, serve pre-compiled versions is a roadblock for would-be contributors. I'm not looking to encourage "drive-by-pull-requests/issues", rather, strive for as pleasant of development experience as possible. I think this would avoid the .whl issues you mentioned.

I've suggested the project re-consider's it's static CSS build system. This not only alleviates packaging issues (for the project and users having to install system packaged), (this is the best part) it also makes SCSS development a breeze. 

It's my opinion that the project would benefit from encouraging/attracting front-end devs. Adding functionality, polish and compatibility with a wide array of frameworks.

I've provided a working prototype, with bootstrap 5 and a new logo as a bonus. Please clone it and let me know what you think. The design / CSS stuff is a bit out of my wheelhouse, I'm primarily a backend developer that pretends to know  DOM/design/etc. :D

benjaoming:
Is there any way for django-compressor's offline compression method to be used for builds? So that builds can have static CSS bundled, rather than depending on django-compressor and libscss.

rsp2k:
@benjaoming https://django-compressor.readthedocs.io/en/stable/usage.html#offline-compression

benjaoming:
Will try to get this in at some point!

Regarding this:

> By default, when a request is made, django-compressor launches libsass as a JIT compiler and returns CSS in the response.

Does django-libsass require any binaries on the host system?

benjaoming:
Just reading https://lukeplant.me.uk/blog/posts/django-sass-scss-without-nodejs-or-build-step/and I understand libsass doesn't have any system requirements :+1:

PR #1389, opened 2025-01-29
Fix history info on logged IPs for anonymous users

muellermartin:
Currently, if the setting `LOG_IPS_ANONYMOUS` is set to `False` no IPs are logged for anonymous users (as expected). But the article history still shows "by anonymous (IP logged)". Instead, the entry should show "by anonymous (IP not logged)".

This PR adds an additional check in the template if an IP address is set in the revision and thus enables the right branch in the template.

(Side note: This is my first PR with absolutely no prior knowledge of this code base so there might be a better way, but I tried my best!)

### DIFF
--- a/src/wiki/templates/wiki/includes/revision_info.html
+++ b/src/wiki/templates/wiki/includes/revision_info.html
@@ -7,7 +7,7 @@
 
 
 {% load wiki_tags i18n %}
-{% if not hidedate %}{{ revision.created }}{% endif %} {% if not hidenumber %}(#{{ revision.revision_number }}) {% trans "by" %}{% endif %} {% if revision.user %}{{ revision.user }}{% else %}{% if article|can_moderate:user %}{{ revision.ip_address|default:"anonymous (IP not logged)" }}{% else %}{% trans "anonymous (IP logged)" %}{% endif %}{% endif %}
+{% if not hidedate %}{{ revision.created }}{% endif %} {% if not hidenumber %}(#{{ revision.revision_number }}) {% trans "by" %}{% endif %} {% if revision.user %}{{ revision.user }}{% else %}{% if article|can_moderate:user or not revision.ip_address %}{{ revision.ip_address|default:"anonymous (IP not logged)" }}{% else %}{% trans "anonymous (IP logged)" %}{% endif %}{% endif %}
 {% if revision == current_revision %}
   <strong>*</strong>
 {% endif %}


src/wiki/templates/wiki/includes/revision_info.html
```diff
@@ -7,7 +7,7 @@
 
 
 {% load wiki_tags i18n %}
-{% if not hidedate %}{{ revision.created }}{% endif %} {% if not hidenumber %}(#{{ revision.revision_number }}) {% trans "by" %}{% endif %} {% if revision.user %}{{ revision.user }}{% else %}{% if article|can_moderate:user %}{{ revision.ip_address|default:"anonymous (IP not logged)" }}{% else %}{% trans "anonymous (IP logged)" %}{% endif %}{% endif %}
+{% if not hidedate %}{{ revision.created }}{% endif %} {% if not hidenumber %}(#{{ revision.revision_number }}) {% trans "by" %}{% endif %} {% if revision.user %}{{ revision.user }}{% else %}{% if article|can_moderate:user or not revision.ip_address %}{{ revision.ip_address|default:"anonymous (IP not logged)" }}{% else %}{% trans "anonymous (IP logged)" %}{% endif %}{% endif %}
```
benjaoming, line 10:
```suggestion
{% if not hidedate %}{{ revision.created }}{% endif %}

{% if not hidenumber %}(#{{ revision.revision_number }}) {% trans "by" %}{% endif %}

{% if revision.user %}
  {{ revision.user }}
{% else %}
{% if not revision.ip_address %}
  {% trans "anonymous (IP not logged)" %}
{% else %}
  {% if article|can_moderate:user %}
    {{ revision.ip_address }}
  {% else %}
    {% trans "anonymous (IP logged)" %}
  {% endif %}
{% endif %}
```

This is admittedly a very unreadable line and I think that since it's HTML we're allowed to add linebreaks without changing the final result :)

We can also shuffle around the logic a bit.

If you like it, can you check and see?

benjaoming (commented):
Thanks for the contribution and please feel welcome to make all sorts of changes like this :+1: 

I would also encourage to add more readability here, I'm not really sure why linebreaks haven't been used. It's probably a very old piece of code :)

benjaoming:
And thanks for a careful elaboration of the issue! Definitely made it easier to jump in and see what's going on :+1:

muellermartin:
Thanks for the feedback! I also like your suggestion to split the template tags into multiple lines for better readability. Usually, additional whitespace should not affect the output, but sometimes it does, e.g. for [inline elements](https://developer.mozilla.org/en-US/docs/Web/API/Document_Object_Model/Whitespace#spaces_in_between_inline_and_inline-block_elements) - which is not the case here.

Previously the whitespace looked like this in the raw HTML:
```
Jan. 29, 2025, 7:49 p.m. (#5) by anonymous (IP not logged)
```

After the line splits it somewhat looks like this:
```
Jan. 29, 2025, 7:49 p.m.
(#5) by

  
    anonymous (IP not logged)
  
```

This might break greppability for the HTML output (e.g. when loading the page via `curl` and searching for log entries via `grep`). Unfortunately, Django's template system doesn't have the cool whitespace control features from Jinja. But I think in this case the code readability is probably more important than "proper" whitespace. Therefore I will go with your suggestion :)

benjaoming:
@muellermartin nice to hear that you have a motivation for keeping it linebreak-free... maybe that's similar to the original intention!

We can wrap it with `{% spaceless %}` in that case. See docs: https://docs.djangoproject.com/en/5.1/ref/templates/builtins/#spaceless

benjaoming:
@muellermartin just wanna say this was a super-great PR and thanks for the final clarification, totally agree!

Sorry for the time it took to get around to this, but happy to merge and release this.

PR #1393, opened 2025-04-24
Add Django 5.2 support

dangrie158:
Here is the followup PR to [django-nyt#146](https://github.com/django-wiki/django-nyt/pull/146) that adds Django 5.2 and Python 3.13 support do django-wiki.


dangrie158:
Ok, the test are not only failing on my machine :)

Seems like `sorl` dropped support for Django 3.X with version 12.11 (latest) (https://pypi.org/project/sorl-thumbnail/ and https://www.djangoproject.com/download/#supported-versions).

I could either pin the previous version 3.10 which still supports Django 3.2 or remove support for 3.2 from django-wiki. What would you prefer?

This probably affects current installs of django-wiki as well.

oscarmcm:
So Django 3.2 (LTS) ended support as of April 2024, I think the best here is to drop support for Django 3.2

cc @benjaoming

benjaoming:
Sorry about the late response - agreed!

benjaoming:
Happy to release this as 0.12 and dropping Django 3.2 support :clap: 

Super-nice work @dangrie158 - thanks for preparing so much of the release and putting in thought of how to go about with it :+1: :heart:

PR #1397, opened 2025-05-29
Update test_views.py

### LINKED ISSUE #1118: Permission tests
After going over testing requirements for a new contribution in #1110, it came up that there is a lack of tests to ensure permissions are upheld. For instance to assert that users without view access to an article also cannot view that article.

These can both be implemented in a separate test module or added as new methods in current view tests.

DankCoder20:
Fixes #1118
Wrote a test to check the given issue .


benjaoming:
Anyone who wants to finish this work should feel welcome!

PR #1404, opened 2025-10-06
Another instance of using safe incorrectly

ryan-esty:
If someone managed to get something malicious into the DB we display it in the template. Similar issue to https://github.com/django-wiki/django-wiki/security/advisories/GHSA-f8hx-4ghf-5wg2


oscarmcm (approved):

benjaoming:
Released as 0.12.2 :+1:

PR #1422, opened 2026-06-23
Add python 3.14 and Django 6.0 to test matrix

benjaoming:


benjaoming:
It seems test failures are now related to issues with codehilite. Will have to investigate some other time.

PetrDlouhy:
I dug into the CI failures here, because I'd like to see a release with the relaxed `pymdown-extensions` pin. Good news: **Django 6.0 and Python 3.14 are fine.** All nine jobs fail with exactly the same two tests and 289 passes — from Python 3.10/Django 4.2 all the way to Python 3.14/Django 6.0. The failure on the pre-existing 3.10/4.2 combination is the giveaway: it's not a compatibility problem.

## Root cause: the pytest bump pulls in Pygments

1. Pygments isn't declared anywhere in `pyproject.toml` — not in project deps, not in the test env.
2. In pytest ≤ 8.3 `pygments` is only an extra (`extra == "testing"` / `"dev"`). In **pytest 8.4.0+ it is an unconditional runtime dependency**. This PR moves `pytest>=6.2.5,<7.3` → `>=9.1,<9.2`.
3. So on `main` the test env has **no Pygments at all**, and `CodehiliteTests` silently takes its `else` (no-Pygments) branch. I confirmed locally: with Pygments uninstalled, the rendered output matches those `else` expectations byte-for-byte.
4. With pytest 9, Pygments comes in transitively and the `if pygments` branches run **for the first time in years** — and their expected HTML is stale by several Pygments releases.

The drift is deep enough to confirm those assertions have never really run in CI: current output adds the `<code>` wrapper (Pygments' `wrapcode`), wraps the table in `<div class="codehilite">`, wraps line numbers in `<span class="normal">`, and classifies `print` as a builtin (`nb`) instead of a keyword (`k`).

Build numbers back this up: `main` (#9224) ran 14 minutes before this PR (#9261) on the same day, same dependency landscape — 291 passed/3 skipped vs 289 passed/2 failed/3 skipped. Same 292 tests, with two flipping from a trivially-passing branch to a real one.

## Suggested fix

Two parts:

- Add `pygments` explicitly to the test env dependencies, so this path is tested on purpose rather than by accident of pytest's dependency graph.
- Update the two expected strings in `tests/core/test_markdown.py`.

Reproduced locally with Markdown 3.9 + Pygments 2.20.0:

<details>
<summary>Corrected expectations (the <code>if pygments</code> branches)</summary>

`test_fenced_code`:

```python
                """<p>Code:</p>\n"""
                """<div class="codehilite-wrap"><div class="codehilite"><pre><span></span><code><span class="n">echo</span> <span class="s1">&#39;line 1&#39;</span>\n"""
                """<span class="n">echo</span> <span class="s1">&#39;line 2&#39;</span>\n"""
                """</code></pre></div>\n"""
                """</div>"""
```

`test_indented_code`:

```python
                """<p>Code:</p>\n"""
                """<div class="codehilite-wrap"><div class="codehilite"><table class="codehilitetable"><tr><td class="linenos"><div class="linenodiv"><pre><span class="normal">1</span>\n"""
                """<span class="normal">2</span>\n"""
                """<span class="normal">3</span>\n"""
                """<span class="normal">4</span></pre></div></td><td class="code"><div><pre><span></span><code><span class="ch">#!/usr/bin/python</span>\n"""
                """<span class="nb">print</span><span class="p">(</span><span class="s1">&#39;line 1&#39;</span><span class="p">)</span>\n"""
                """<span class="nb">print</span><span class="p">(</span><span class="s1">&#39;line 2&#39;</span><span class="p">)</span>\n"""
                """<span class="nb">print</span><span class="p">(</span><span class="s1">&#39;æøå&#39;</span><span class="p">)</span>\n"""
                """</code></pre></div></td></tr></table></div>\n"""
                """</div>"""
```

</details>

One caveat: exact-HTML assertions against Pygments will keep rotting like this. Either pin Pygments in the test env, or assert on structure instead of an exact string — your call.

If you just want to unblock the release, the pytest bump is orthogonal to the Django 6.0 / Python 3.14 support and could be split out, leaving this PR green immediately.

Happy to open a PR with the fix if that helps.

PetrDlouhy:
I've opened #1425 with the fix: declare `pygments` in the hatch env dependencies, and refresh the two stale expectations.

It's independent of this PR and targets `main`, so once it lands, the pytest bump here no longer changes anything about those two tests. Verified green locally under both pytest 7.2.2 (current pin) and 9.1.1 (the pin proposed here).

benjaoming:
Tested locally quickly and things are looking good!

PR #1425, opened 2026-07-31
Declare Pygments as a test dependency and refresh codehilite expectations

PetrDlouhy:
## Why

`CodehiliteTests` asserts on Pygments-rendered HTML, but **Pygments is declared nowhere** — not in the project requirements, not in the hatch env. Whether it gets installed is therefore decided by pytest's own dependency graph: it is an extra (`extra == "testing"` / `"dev"`) up to pytest 8.3, and an **unconditional runtime dependency from pytest 8.4 onwards**.

With the current `pytest>=6.2.5,<7.3` pin, Pygments is absent from the test env, so both tests silently take their `else` (no-Pygments) branch. The `if pygments` branches have not run in CI for years, and drifted out of date accordingly.

This is latent breakage that surfaces the moment anything pulls Pygments in — which is exactly what happens in #1422, where the bump to pytest 9 makes **all nine matrix jobs** fail on these two tests, while 289 tests pass everywhere else (including Django 6.0 and Python 3.14).

## What changed

- Declare `pygments` in the hatch env dependencies, so the code path is exercised on purpose rather than by accident of pytest's dependency graph.
- Update the two expected strings to current Pygments output. Four separate drifts had accumulated:
  - code wrapped in `<code>` (Pygments' `wrapcode`)
  - the linenos table wrapped in `<div class="codehilite">`
  - line numbers wrapped in `<span class="normal">`
  - `print` classified as a builtin (`nb`) rather than a keyword (`k`)

## Testing

Verified locally with Markdown 3.9 + Pygments 2.20.0, under **both** pytest pins:

| pytest | `tests/core/test_markdown.py` |
|---|---|
| 7.2.2 (current pin) | 5 passed |
| 9.1.1 (pin proposed in #1422) | 5 passed |

Full `tests/core/` on the current pin: **135 passed, 3 skipped**.

## Scope / follow-ups

Deliberately *not* changed here: the `else` (no-Pygments) branches, which still pass and are left as-is.

Asserting on exact Pygments HTML stays brittle across Pygments releases — this will rot again. Pinning Pygments in the test env, or asserting on structure rather than an exact string, would be a sensible follow-up, but felt out of scope for an unblocking fix.

This does not depend on #1422 and unblocks it: once merged, that PR's pytest bump no longer changes anything about these tests.

### DIFF
--- a/testproject/testproject/settings/dev.py
+++ b/testproject/testproject/settings/dev.py
@@ -20,7 +20,7 @@ ALLOWED_HOSTS = [
 # Removed.
 # See: https://forum.djangoproject.com/t/why-are-cookie-secure-settings-defaulted-to-false/1133/4
 # and https://github.com/django-wiki/django-wiki/pull/1325
-# SESSION_COOKIE_DOMAIN = ".localhost"
+SESSION_COOKIE_DOMAIN = None
 SESSION_COOKIE_SECURE = False
 
 try:
--- a/docs/tips/index.rst
+++ b/docs/tips/index.rst
@@ -23,7 +23,6 @@ Quick tips
            'extensions': [
                'footnotes',
                'attr_list',
-               'headerid',
                'extra',
                'codehilite',
            ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -141,9 +141,7 @@ dependencies = [
     "pytest-pythonpath",
     "pytest>=6.2.5,<7.3",
     # Pygments is optional for django-wiki itself, but the codehilite tests
-    # assert on its output. Without it declared here, whether it is installed
-    # depends on pytest's own dependencies (unconditional since pytest 8.4),
-    # and the Pygments code paths silently go untested.
+    # assert on its output.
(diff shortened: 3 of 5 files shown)


benjaoming:
Tried it out locally, had some other adjacent issues with newer Django versions and amending changes here (just because this is the PR I'm testing and focusing on now)

testproject/testproject/settings/dev.py
```diff
@@ -20,7 +20,7 @@
 # Removed.
 # See: https://forum.djangoproject.com/t/why-are-cookie-secure-settings-defaulted-to-false/1133/4
 # and https://github.com/django-wiki/django-wiki/pull/1325
-# SESSION_COOKIE_DOMAIN = ".localhost"
+SESSION_COOKIE_DOMAIN = None
```
benjaoming, line 23: This was unrelated to the fix, but necessary to login on the testproject and try things out.

benjaoming (approved):

